PR #31, opened 2022-02-07
Fill the protocol info for gRPC request in service and topology metric

NeJan2020:
We need requestContent and responseContent to show requestPath and responseHttpStatus of gRPC request (just like http request).
lt should be marged in uprobeAnalyzer and relabeled in formathProcessor, but doesn't work as expected now .

### DIFF
--- a/collector/consumer/processor/kindlingformatprocessor/protocol.go
+++ b/collector/consumer/processor/kindlingformatprocessor/protocol.go
@@ -56,7 +56,7 @@ func fillCommonProtocolLabels(g *gauges, protocol ProtocolType, isServer bool) {
 		}
 	case grpc:
 		if isServer {
-			fillEntityGRPCProtocolLabel(g)
+			fillEntityHttpProtocolLabel(g)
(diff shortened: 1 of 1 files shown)


collector/analyzer/uprobeanalyzer/uprobeanalyzer.go
```diff
@@ -11,8 +12,11 @@ import (
 	"github.com/Kindling-project/kindling/collector/model/constvalues"
 	"github.com/hashicorp/go-multierror"
 	"go.uber.org/zap"
+	"strings"
 )
 
+var httpMerger = tools.NewHttpMergeCache()
```
dxsup, line 18: It seems like tools.httpMerger is used in two different packages now, then we should remove this tool from `networkanalyzer` and` put it into a new package called `analyzerhelper` under package `analyzer` to decouple networkanalyzer from uprobeanalyzer.

dxsup (changes requested):

PR #32, opened 2022-02-10
docs: fix installation guide location

yowenter:


dxsup:
Thanks for the PR, but we'll change the docs path to our official websites, so this PR will be closed after the modification is finished.

PR #37, opened 2022-02-14
Fix/fill the protocol info for grpc request in service and topology metric

NeJan2020:
We need requestContent and responseContent to show requestPath and responseHttpStatus of gRPC request (just like http request).
lt should be marged in uprobeAnalyzer and relabeled in formathProcessor, but doesn't work as expected now .


dxsup:
LGTM

PR #38, opened 2022-02-15
The collector judges whether the socket with the probe is normal according to the status code of zeromq

### LINKED ISSUE #34: Can't receive events after the probe is restarted
**Describe the bug**
Can't receive events after the probe is restarted.
**What you expected to happen**
The collector is supposed to receive events from the probe.
**How to reproduce**
Start kindling and restart the probe, then you will see the collector can't receive any events.
**Additional context**
The collector starts to receive events after sending the subscribing events. The root cause of this bug is that after the probe is restarted, it loses information about subscribing in which case it won't send events.
Therefore, to fix this bug, the feature that the collector can resend the subscribing events is requested.

jundizhou:
fix #34 
when pull event faild in three times,restart connect


jundizhou:
This is not a good solution. So we chose to close it temporarily

PR #45, opened 2022-02-17
Fix bug that always return 0 when the input string is in IPv4 format

### LINKED ISSUE #44: Can't get network destination pod name from network dashbord
**Describe the bug**
Hi, Kindling guys . In network details dashboard I found my pods' connections  encounter  pkg retransmit, but when i want to know connection's pod details  ,i get the source pod name from the <retransmit tables> while destination pod name is missing. By the way, the network destination must be a pod , cause i get the destination workload name.

**What did you expect to see?**
i want to get the network connection's source pod name and also destination pod name.

**What did you see instead?**
A clear and concise description of what you saw instead.

**Screenshots**
image

dxsup:
```go
func StringToUint32(ip string) uint32 {
	bytes := strings.Split(ip, ".")
        // This line will make the result always be zero when the input IP string is in IPv4 fomat
	if len(bytes) <= 4 {
		return 0
	}
	b0, _ := strconv.Atoi(bytes[0])
	b1, _ := strconv.Atoi(bytes[1])
	b2, _ := strconv.Atoi(bytes[2])
	b3, _ := strconv.Atoi(bytes[3])

	return uint32(b0) | uint32(b1)<<8 | uint32(b2)<<16 | uint32(b3)<<24
}
```
fix #44

### DIFF
--- a/collector/metadata/conntracker/utils.go
+++ b/collector/metadata/conntracker/utils.go
@@ -17,7 +17,7 @@ func IPToUInt32(ip net.IP) uint32 {
 
 func StringToUint32(ip string) uint32 {
 	bytes := strings.Split(ip, ".")
-	if len(bytes) <= 4 {
+	if len(bytes) < 4 {
 		return 0
 	}
 	b0, _ := strconv.Atoi(bytes[0])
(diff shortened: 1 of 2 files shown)


NeJan2020 (approved):
LGTM

PR #47, opened 2022-02-17
Fix Parse kafaka request failed by index out of range

hocktea214:
The clientLength is negative that caused index out of range.


dxsup (commented):
Is there also the same problem with`func (message *PayloadMessage) ReadInt64(offset int)`? Could this problem be  triggered by other functions?

hocktea214:
Other functions have been checked and fixed. Methods like ReadInt64() which is not used is now removed.

dxsup (approved):

PR #48, opened 2022-02-18
Store the topology metric whose src_ip is from external namespace

### LINKED ISSUE #46: The external network traffics are not being traced
**Describe the bug**
Hi, I want to identify the specific source that leads to the heavy traffics on my ingress workload. But in the  topology dashboard, i can't see the external IP  to my ingress workload, it seems that the  external network traffics are not being traced 
**How to reproduce?**
Visit your ingress from outside the Kubernetes cluster

**What did you expect to see?**
External IP to ingress must be displayed

**What did you see instead?**
No node sends traffic to my ingress workload
**Screenshots**
image

dxsup:
Generate a topology gauge when the data from the server-side contains an external src IP.

This PR will close #46

### DIFF
--- a/collector/application/testdata/kindling-collector-config.yaml
+++ b/collector/application/testdata/kindling-collector-config.yaml
@@ -60,6 +60,7 @@ processors:
   kindlingformatprocessor:
     need_trace_as_metric: true
     need_pod_detail: true
+    store_external_src_ip: true
   nodemetricprocessor:
 
 exporters:
--- a/collector/consumer/processor/k8sprocessor/kubernetes_processor.go
+++ b/collector/consumer/processor/k8sprocessor/kubernetes_processor.go
@@ -95,7 +95,7 @@ func (p *K8sMetadataProcessor) addK8sMetaDataForClientLabel(labelMap *model.Attr
 	} else {
 		labelMap.AddStringValue(constlabels.SrcNodeIp, p.localNodeIp)
 		labelMap.AddStringValue(constlabels.SrcNode, p.localNodeName)
-		labelMap.AddStringValue(constlabels.SrcNamespace, constlabels.ExternalClusterNamespace)
+		labelMap.AddStringValue(constlabels.SrcNamespace, constlabels.InternalClusterNamespace)
 	}
 	// add metadata for dst
 	dstIp := labelMap.GetStringValue(constlabels.DstIp)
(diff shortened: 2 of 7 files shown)


NeJan2020 (approved):
LGTM

dxsup:
There are still some problems to fix for this PR. I'll re-request review once they are fixed.---- Replied Message ***@***.***>Date02/21/2022 09:21 ***@***.***> CcDaxin ***@***.******@***.***>SubjectRe: [Kindling-project/kindling] Store the topology metric whose src_ip is from external namespace (PR #48)
@NeJan2020 approved this pull request.

LGTM

—Reply to this email directly, view it on GitHub, or unsubscribe.Triage notifications on the go with GitHub Mobile for iOS or Android.
You are receiving this because you authored the thread.Message ID: ***@***.***>
[
{
***@***.***": "http://schema.org",
***@***.***": "EmailMessage",
"potentialAction": {
***@***.***": "ViewAction",
"target": "https://github.com/Kindling-project/kindling/pull/48#pullrequestreview-888123603",
"url": "https://github.com/Kindling-project/kindling/pull/48#pullrequestreview-888123603",
"name": "View Pull Request"
},
"description": "View this Pull Request on GitHub",
"publisher": {
***@***.***": "Organization",
"name": "GitHub",
"url": "https://github.com"
}
}
]
        
    
        font{
            line-height: 1.6;
        }

dxsup:
The current implementation causes the topology metric becomes to be double when the `src_ip` is an internal IP but isn't a Pod IP, like NodeIP, in which case both the client-side and the server-side data are stored.
We can't just throw away the external client-side data to solve this problem because we want the data holding "EXTERNAL" both at `src_namesapce` and `dst_namespace` fields, which the server-side one can't provide. The root cause of this problem is that the value "EXTERNAL" at the `namespace` field is unclear as it holds two meanings. One for "the real external namespace outside the cluster" and another one for "metadata not found but inside the cluster". 
The problem can be solved if we name the first meaning above as "NOT_FOUND_EXTERNAL" and name the second one as "NOT_FOUND_LOCAL".

NeJan2020:
In that case, only the client-side data with a `src_ip` which is equal to one of the `node_ip` in the cluster should use 'NOT_FOUND_LOCAL' as its `src_namespace`

dxsup:
image
This is the result we got after changing for topology metric. And the value is as expected with the entity metric below.
image
image
Note they have different values for field `request_content`.

NeJan2020 (approved):
LGTM, but this change will modify the expected output data ,need to be sync with Grafana dashboard

dxsup:
> LGTM, but this change will modify the expected output data ,need to be sync with Grafana dashboard

Right. I'll ask @thousandxu to commit changes needed according to the latest data.

Issue #54, opened 2022-02-22
What are the minimal set of Linux capabilities the agent requests?


dxsup:
There are some security concerns regarding why the agent must use root privileges. It's typical that the users only want to give the minimal set of privileges if possible. We need to explain which privileges are necessary and why. Then we could set Linux capabilities to restrict the range of privileges.

https://github.com/Kindling-project/kindling/blob/fe3467ba0192895762ec2d9eeb89c5bd712de160/deploy/kindling-deploy.yml#L56

https://github.com/Kindling-project/kindling/blob/fe3467ba0192895762ec2d9eeb89c5bd712de160/deploy/kindling-deploy.yml#L102

sanyangji:
[...]

PR #56, opened 2022-02-22
use minimum capabilities for collector container

dxsup:


dxsup:
Related to issue #54

Issue #63, opened 2022-03-02
Performance tuning for the collector


dxsup:
We have conducted a lot of benchmarks for the collector and found several essential points which cause a great impact on the performance. This issue is created to record the process of tuning.

dxsup:
Testing was completed on CentOS7 whose kernel version is `3.10.0-1127.el7.x86_64`. There are 8 CPUs(2.5GHz) and 16GB of memory.

First, we ran a load test where there is an Nginx running under 5k QPS and the Nginx sent a request to a backend service for every request received. So there are a total of 10k QPS in this node and 25k+ events received per second. The command `top` showed the collector used 125% CPU in this scenario which is not affordable for production usage.

We used go-torch to generate a flame graph and found that `runtime.gcBgMarkWorker` spent 40%  of the CPU time. It’s one of the methods called by the garbage collector.
![torch-no-zap](https://user-images.githubusercontent.com/46807570/156291392-e0f52052-aba2-4cd1-ba69-63bb53e4b85e.svg)

The performance of the garbage collector is affected by how many objects are allocated and how frequently the collection is executed. To figure out why the garbage collector spends so much CPU time, we dumped the heap using `go tool pprof` to find which and where the objects are allocated. 

From the result, we could know there are fives parts allocating the most objects.
- tools.HttpMerger.GetContentKey(23%)
![image](https://user-images.githubusercontent.com/46807570/156291767-ae0ae76d-f18a-453f-9d62-157396179a89.png)
- otelexporter (16.35%)
![image](https://user-images.githubusercontent.com/46807570/156291863-22481567-a4d9-449d-bfb3-49e6cf08f863.png)
- UdsReceiver.unmarshal(14.42%)
![image](https://user-images.githubusercontent.com/46807570/156291679-7a3c4eab-265b-4e9b-8de3-767f320fa627.png)
- network.*NetworkAnalyzer.getRecords(11.84%)
![image](https://user-images.githubusercontent.com/46807570/156291894-19344966-e03c-4af9-b14c-1841d3ebef23.png)
- kindlingformatprocessor.gauges.Process(8.72%)
![image](https://user-images.githubusercontent.com/46807570/156291931-a5b0f451-4057-4e45-a750-ccd5f8534e8d.png)

dxsup:
## tools.HttpMerger.GetContentKey(23%)
This part is out of our expectations. The root cause is that `regexp.MatchString` re-compiles the same regular expression every time. 
```go
[...]

PR #65, opened 2022-03-02
Use sync.Pool to reuse objects when networkanalyzer generates a new request

dxsup:
Related to issue #63 

The `GaugeGroup` object is put back to the pool once it passes through the processing chain. Note we reuse the objects created before when the `NetworkAnalyzer` generates a new request, and all the labels added before are still there, so there may be some unuseful labels that have to be discarded by later processors.

We also add an `Update` method to `attributeMap` to make the labels reusable.

### DIFF
--- a/collector/analyzer/network/gauge_pool.go
+++ b/collector/analyzer/network/gauge_pool.go
@@ -1,0 +1,39 @@
+package network
+
+import (
+	"github.com/Kindling-project/kindling/collector/model"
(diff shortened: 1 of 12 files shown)


dxsup:
> LGTM,Maybe we can also use sync.pool in processor and exporter later

Right! I have written that down in https://github.com/Kindling-project/kindling/issues/63#issuecomment-1056567029. We could track the progress on it.

PR #66, opened 2022-03-03
Add variable GOGC=400 to the deployment yaml

dxsup:
Related to issue #63.
Like what we found https://github.com/Kindling-project/kindling/issues/63#issuecomment-1056647203, we could set GOGC  400 to reduce the frequency of GC.

### DIFF
--- a/deploy/kindling-deploy.yml
+++ b/deploy/kindling-deploy.yml
@@ -107,6 +107,8 @@ spec:
           requests:
             memory: 300Mi
         env:
+        - name: GOGC
+          value: "400"
         - name: MY_NODE_IP
           valueFrom:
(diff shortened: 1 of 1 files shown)


NeJan2020 (approved):
It is not a fundamental solution to the problem, but it does improve the performance of CPU Usage by reduced GC

dxsup:
> It is not a fundamental solution to the problem, but it does improve the performance of CPU Usage by reduced GC

Right, this change will increase memory usage. I have written down what should we do later to reduce the allocation of objects at https://github.com/Kindling-project/kindling/issues/63#issuecomment-1056567029, which will leverage the garbage collector fundamentally.

PR #68, opened 2022-03-03
Do not use srcPort as a part of key when the event is with TCP type

dxsup:
Related to issue #63.
See more https://github.com/Kindling-project/kindling/issues/63#issuecomment-1057582803.


hocktea214 (approved):
Udp protocol such as DNS will check ip ports , tcp just check fd .

PR #69, opened 2022-03-03
filter out events belonging to container runtime daemon

jundizhou:
For the time being, we can filter out these events in the user mode. Later, we can consider filtering in the kernel


sanyangji:
The way looks not graceful temporally, but to some extent, it can filter out some events we don't concern and the events will not be sent to Collector.
Actually You may call function `bool sinsp::suppress_events_comm(const std::string &comm)` in agent-libs to suppress the process in libscap.
It's better to add one controller to control the filter in kernel in the future.

sanyangji (approved):
The optimization work should be done later.

PR #70, opened 2022-03-03
Add metadata of container to the output data

dxsup:
Since we can get the container metadata, the output metrics and traces could also include it. It is useful when the users want to know which container receives or sends the requests.

The cardinality of metric doesn't increase much because a combination of `Port` and `PodName` already included before actually stands for exact one container/container's id.


NeJan2020 (approved):
LGTM

PR #71, opened 2022-03-03
Rename protocol "generic" to "NOSUPPORT" to make it more comprehensible

dxsup:
Rename protocol name “generic” to “UNSUPPORTED” to make it more comprehensible. Note the metric with this protocol is not accurate.

### DIFF
--- a/collector/analyzer/network/network_analyzer.go
+++ b/collector/analyzer/network/network_analyzer.go
@@ -319,7 +319,7 @@ func (na *NetworkAnalyzer) parseProtocols(mps *messagePairs) []*model.GaugeGroup
 		for _, parser := range cacheParsers {
 			records := na.parseProtocol(mps, parser)
 			if records != nil {
-				if protocol.UNSUPPORTED == parser.GetProtocol() {
+				if protocol.NOSUPPORT == parser.GetProtocol() {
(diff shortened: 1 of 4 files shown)


collector/analyzer/network/network_analyzer.go
```diff
@@ -319,7 +319,7 @@ func (na *NetworkAnalyzer) parseProtocols(mps *messagePairs) []*model.GaugeGroup
 		for _, parser := range cacheParsers {
 			records := na.parseProtocol(mps, parser)
 			if records != nil {
-				if protocol.GENERIC == parser.GetProtocol() {
+				if protocol.UNSUPPORTED == parser.GetProtocol() {
 					// Reset mapping for  generic and port when exceed threshold so as to parsed by other protcols.
```
hocktea214, line 323: The comment with generic need to rename.

hocktea214 (dismissed):
LGTM

dxsup:
We talked about which word should be used here. There are three options: `UNSUPPORT`, `UNSUPPORTED` or `NOSUPPORT`. We think the field `protocol` should hold a noun. Considering that `UNSUPPORT` is not a valid word, we prefer to choose `NOSUPPORT` as the final option.

hocktea214 (approved):

PR #78, opened 2022-03-03
Fix the trace data model not consistent with the older version

dxsup:


NeJan2020 (approved):
LGTM

Issue #80, opened 2022-03-04
lower kernel version not support metric of srtt


jundizhou:
**Describe the bug**
lower kernel version not support metric of srtt

**How to reproduce?**
compiler probe in 3.10.0-229.el7.x86_64 && 3.10.0-123.el7.x86_64

**Logs**
Please attach the logs by running the following command:
```
kubectl logs -f kindling-agent-xxx(replace with your podname) -n kindling -c kindling-probe
kubectl logs -f kindling-agent-xxx(replace with your podname) -n kindling -c kindling-collector
```

14:42
Compile probe for 3.10.0-123.el7.x86_64
make -C /usr/src/kernels/3.10.0-123.el7.x86_64/ M=/source/driver modules
make[1]: Entering directory `/host/usr/src/kernels/3.10.0-123.el7.x86_64'
  CC [M]  /source/driver/main.o
  CC [M]  /source/driver/dynamic_params_table.o
  CC [M]  /source/driver/fillers_table.o
  CC [M]  /source/driver/flags_table.o
  CC [M]  /source/driver/ppm_events.o
  CC [M]  /source/driver/ppm_fillers.o
/source/driver/ppm_fillers.c: In function 'f_tcp_rcv_established_e':
/source/driver/ppm_fillers.c:5224:15: error: 'struct tcp_sock' has no member named 'srtt_us'
  u32 srtt = ts->srtt_us >> 3;
               ^
make[2]: *** [/source/driver/ppm_fillers.o] Error 1
make[1]: *** [_module_/source/driver] Error 2
make[1]: Leaving directory `/host/usr/src/kernels/3.10.0-123.el7.x86_64'
make: *** [all] Error 2
mv: cannot stat 'kindling-falcolib-probe.ko': No such file or directory
make -C /usr/src/kernels/3.10.0-123.el7.x86_64/ M=/source/driver clean
make[1]: Entering directory `/host/usr/src/kernels/3.10.0-123.el7.x86_64'
  CLEAN   /source/driver/.tmp_versions
make[1]: Leaving directory `/host/usr/src/kernels/3.10.0-123.el7.x86_64'

jundizhou:
I will add judgment for kernel version and special judgment for CentOS
[...]

Issue #60, opened 2022-03-01
When compiling probe, ebpf modules should not be compiled by lower versions


jundizhou:
**Describe the bug**
When compiling the probe, only the kernel module will succeed in the low version kernel, and ebpf will fail, but it will not affect the use We need to optimize this. Ebpf will not be compiled in the lower version

**What did you expect to see?**
compiler success in lower version,not so much error log

**Logs**
In file included from include/linux/sched.h:33:
include/linux/signal.h:220:10: warning: array index 1 is past the end of the array (which contains 1 element) [-Warray-bounds]
        case 2: set->sig[1] = 0;
                ^        ~
./arch/x86/include/asm/signal.h:23:2: note: array 'sig' declared here
        unsigned long sig[_NSIG_WORDS];
        ^
In file included from /source/driver/bpf/probe.c:16:
In file included from include/linux/sched.h:33:
include/linux/signal.h:232:10: warning: array index 1 is past the end of the array (which contains 1 element) [-Warray-bounds]
        case 2: set->sig[1] = -1;
                ^        ~
./arch/x86/include/asm/signal.h:23:2: note: array 'sig' declared here
        unsigned long sig[_NSIG_WORDS];
        ^
In file included from /source/driver/bpf/probe.c:20:
/source/driver/bpf/bpf_helpers.h:13:10: error: use of undeclared identifier 'BPF_FUNC_map_lookup_elem'
        (void *)BPF_FUNC_map_lookup_elem;
                ^
/source/driver/bpf/bpf_helpers.h:16:10: error: use of undeclared identifier 'BPF_FUNC_map_update_elem'
        (void *)BPF_FUNC_map_update_elem;
                ^
/source/driver/bpf/bpf_helpers.h:18:10: error: use of undeclared identifier 'BPF_FUNC_map_delete_elem'
        (void *)BPF_FUNC_map_delete_elem;
                ^
/source/driver/bpf/bpf_helpers.h:20:10: error: use of undeclared identifier 'BPF_FUNC_probe_read'
        (void *)BPF_FUNC_probe_read;
                ^
/source/driver/bpf/bpf_helpers.h:22:10: error: use of undeclared identifier 'BPF_FUNC_ktime_get_ns'
        (void *)BPF_FUNC_ktime_get_ns;
                ^
/source/driver/bpf/bpf_helpers.h:24:10: error: use of undeclared identifier 'BPF_FUNC_trace_printk'
        (void *)BPF_FUNC_trace_printk;
[...]

PR #97, opened 2022-03-10
wip:support srtt on older versions of CentOS & skip compiling eBPF probes on older versions

jundizhou:
#60 #80

### DIFF
--- a/probe/WORKSPACE
+++ b/probe/WORKSPACE
@@ -22,9 +22,9 @@ http_archive(
 
 http_archive(
         name = "agent-libs",
-        urls = ["https://github.com/Kindling-project/agent-libs/archive/8c130f7c4326398372661c0148b5bba5a1a2e9d8.tar.gz"],
-        sha256 = "08eeda3700337e7ce0bc69eb8b0634266935923764b6494f1cb52eb5dc4d334c",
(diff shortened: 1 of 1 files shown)


sanyangji:
Due to this commit is based on Kindling-project/agent-libs, I checked the commit in Kindling-project/agent-libs. The change on `/usr/src/kernels` will fail the script on Debian-like systems because they install the headers under `/usr/src` directly without the directory `kernels`.
Besides, two more improvements for the script: 1. The judge logic can be compressed for compiling eBPF probes. 2. The logic of compiling kernel modules can be separated too.

jundizhou:
> Due to this commit is based on Kindling-project/agent-libs, I checked the commit in Kindling-project/agent-libs. The change on `/usr/src/kernels` will fail the script on Debian-like systems because they install the headers under `/usr/src` directly without the directory `kernels`. Besides, two more improvements for the script: 1. The judge logic can be compressed for compiling eBPF probes. 2. The logic of compiling kernel modules can be separated too.

debian’s question cloud create another question.I will do , two improvements for the script

jundizhou:
Abandoned pr，new pr is https://github.com/Kindling-project/kindling/pull/129

PR #98, opened 2022-03-10
Update installation.md

sugary199:
1、change the command of cloning branch code
2、The copy operation usually reports an error because the previous absolute path is different  
3、There is also a scripts folder below the agent-lib directory，we need go back to probe directory


sugary199:
Signed-off-by: yushu <906940958@qq.com>

Issue #96, opened 2022-03-10
modify the command of cloning agent-libs


sugary199:
When I followed the instructions to compile probe, I found that I couldn't pull the branch code
[installation/Installation.md      Build local kernel modules and eBPF modules](https://github.com/Kindling-project/kindling/blob/main/installation/Installation.md#build-local-kernel-modules-and-ebpf-modules)

`git clone https://github.com/Kindling-project/agent-libs/commits/kindling-dev`

it works if we replace it with the following command
`git clone -b kindling-dev https://github.com/Kindling-project/agent-libs`

PR #102, opened 2022-03-10
update Installation.md

sugary199:
Signed-off-by: sugary199 <906940958@qq.com>


sugary199:
1、change the command of cloning branch code
2、The copy operation usually reports an error because the previous absolute path is different
3、There is also a scripts folder below the agent-lib directory，we need go back to probe directory

dxsup:
Anyway, I think this PR could be merged except for the comment I left above, as it will spend some time to finish the new documentation.

dxsup (changes requested):

dxsup (approved):

dxsup:
Close #96

PR #105, opened 2022-03-11
fix the bug of "collector already started"

### LINKED ISSUE #103: "controller already started" reported frequently
The following logs have been reported recently.
```log
2022/03/11 01:43:36 controller already started
2022/03/11 01:43:50 controller already started
2022/03/11 01:43:51 controller already started
2022/03/11 01:44:05 controller already started
2022/03/11 01:44:06 controller already started
2022/03/11 01:44:20 controller already started
2022/03/11 01:44:21 controller already started
2022/03/11 01:44:35 controller already started
2022/03/11 01:44:36 controller already started
2022/03/11 01:44:50 controller already started
2022/03/11 01:44:51 controller already started
2022/03/11 01:45:05 controller already started
2022/03/11 01:45:06 controller already started
2022/03/11 01:45:20 controller already started
2022/03/11 01:45:21 controller already started
2022/03/11 01:45:35 controller already started
2022/03/11 01:45:36 controller already started
```
This is printed by `opentelemetry-go` only if methods `Start` and `Collect` are both used.
``` go
// ErrControllerStarted indicates that a controller was started more
// than once.
var ErrControllerStarted = fmt.Errorf("controller already started")
...
// Note that it is not necessary to Start a controller when only
// pulling data; use the Collect() and ForEach() methods directly in
// this case.
func (c *Controller) Start(ctx context.Context) error {
...
}
```

I have confirmed that `opentelemetry-go/prometheus-exporter` would use the method `Collect`, so it is unnecessary to use the `Start` method. The following codes cause this bug.
https://github.com/Kindling-project/kindling/blob/c55ce70bb8ab968c9da5873227ef82e6bdd7ef79/collector/observability/telemetry.go#L65-L68

Besides, as we can see, the logs printed are not in the same format as ours. Opentelemetry-go provides a global method to set user-defined error handler which should be used to print more information.

dxsup:
Close #103

### DIFF
--- a/collector/consumer/exporter/otelexporter/otelexporter.go
+++ b/collector/consumer/exporter/otelexporter/otelexporter.go
@@ -130,7 +130,12 @@ func NewExporter(config interface{}, telemetry *component.TelemetryTools) export
 			metricAggregationMap: cfg.MetricAggregationMap,
 			telemetry:            telemetry,
 		}
-		go StartServer(exp, telemetry.Logger, cfg.PromCfg.Port)
+		go func() {
+			err := StartServer(exp, telemetry.Logger, cfg.PromCfg.Port)
+			if err != nil {
+				telemetry.Logger.Warn("error starting otelexporter prometheus server: ", zap.Error(err))
+			}
+		}()
 	} else {
 		var collectPeriod time.Duration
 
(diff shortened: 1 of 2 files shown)


NeJan2020 (approved):
LGTM.  After removing the `c.Start()`, I can still get metrics from the HTTP server.

PR #109, opened 2022-03-14
Retain the direction of tcp metric as it is what we need

dxsup:
Think of the request `A->B`, there are two TCP flows, which are A calls B and B returns to A, associated with it. We tried identifying the TCP flow direction to make it associated with the business request before, such as for the metric of retransmitting packet count, we want to change both the TCP flows of `A->B` and `B->A` to `A->B` so as to know whether the performance of a business request is affected by these two direction flows.

This is designed for simplifying the process of querying. But it turns out that we can not classify these directions accurately, so we decided to retain the real direction of the metric and leave the process of association to querying.


NeJan2020 (approved):
LGTM. By the way, that was strange to use SrcNamespace to show whether the DST is from InternalCluster or ExternalCluster which we did before.

dxsup:
> LGTM. By the way, that was strange to use SrcNamespace to show whether the DST is from InternalCluster or ExternalCluster which we did before.

It's a typo and caused a bug. I have fixed it in this PR.

PR #111, opened 2022-03-15
Add the port label for tcp metrics

dxsup:
Metric shouldn't contain high-cardinality labels, so here we want to remove the dynamic port label and retain the listening one. But we can't know which port is dynamic for sure, so we work around that by comparing their number size.

The default dynamic port range in /proc/sys/net/ipv4/ip_local_port_range is 32768~60999. In most cases, the larger port is the dynamic port and the other one is the listening port. But sometimes the listening port is also greater than 32768 in which case there is no way to tell which one is listening. So we just remove all the ports that are larger than 32768.


NeJan2020 (approved):
LGTM

PR #124, opened 2022-03-18
feat(exporter): add self monitor metrics for otel-exporter

NeJan2020:


dxsup:
I think this file `collector/consumer/exporter/otelexporter/metrics.go` should be renamed since the term `metrics` in the `otel-exporter` module is used to represent the general metrics data. `self_telemetry.go` may be a choice.

dxsup (approved):

PR #130, opened 2022-03-22
Fix sub string operation made string none-utf8

hocktea214:
Signed-off-by: zheng <huxiangyuan@harmonycloud.cn>

### DIFF
--- a/collector/analyzer/tools/utf8.go
+++ b/collector/analyzer/tools/utf8.go
@@ -1,0 +1,167 @@
+package tools
+
+import (
+	"unicode/utf8"
+)
+
+const (
+	// The default lowest and highest continuation byte.
+	locb = 0x80 // 1000 0000
+	hicb = 0xBF // 1011 1111
+
+	// These names of these constants are chosen to give nice alignment in the
+	// table below. The first nibble is an index into acceptRanges or F for
+	// special one-byte cases. The second nibble is the Rune length or the
+	// Status for the special one-byte case.
+	xx = 0xF1 // invalid: size 1
+	as = 0xF0 // ASCII: size 1
+	s1 = 0x02 // accept 0, size 2
+	s2 = 0x13 // accept 1, size 3
+	s3 = 0x03 // accept 0, size 3
+	s4 = 0x23 // accept 2, size 3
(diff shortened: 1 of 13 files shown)


collector/analyzer/tools/utf8.go
```diff
@@ -0,0 +1,167 @@
...
+}
+
+func runeIndex(p []byte) int {
+	np := len(p)
+	var n int
+	for i := 0; i < np; {
+		c := p[i]
+		if c < utf8.RuneSelf {
+			// ASCII fast path
+			if n > 0 && n < 3 {
```
dxsup, line 72: Why is it 3 here? Could you add some comments to state the reason?

hocktea214: As the size of chinese character is 3, we only support it for now.

hocktea214: > Why is it 3 here? Could you add some comments to state the reason?
As the size of chinese character is 3, we only support it for now.

hocktea214: > Why is it 3 here? Could you add some comments to state the reason?

As the size of chinese character is 3, we only support it for now.

collector/analyzer/tools/utf8.go
```diff
@@ -0,0 +1,167 @@
...
+		} else if c := s[i+3]; c < locb || hicb < c {
+			size = 1
+		}
+		n = size
+		i += size
+	}
+	return ns
+}
+
+func FomratStringToUtf8(s string) string {
```
dxsup, line 158: I think `Fomrat` is a typo which is supposed to be `Format`.

dxsup:
Since the part of file `utf8.go` is copied from the standard library `unicode/utf8`, I think we should include the license text at the head of the file.

collector/analyzer/tools/utf8.go
hocktea214: Wrong spell word, check it.

hocktea214: It‘s a spelling mistake, fix it.

dxsup (approved):

PR #150, opened 2022-03-25
Add: stirling control flags && Mod: dev base image

KayzzzZ:
* Add control flag for stirling
* Use GLOG for logs instead of using cout
* Update pixie module

Signed-off-by: qianlu.kk <qianlu.lk@alibaba-inc.com>

### DIFF
--- a/probe/WORKSPACE
+++ b/probe/WORKSPACE
@@ -3,9 +3,8 @@ workspace(name = "kp")
 load("@bazel_tools//tools/build_defs/repo:git.bzl","git_repository")
 git_repository(
     name = "px",
-    remote = "https://github.com/KayzzzZ/pixie.git",
-    commit = "6b853b96987e570e0979620537c23df5850c7b76",
-#    branch = "kindling",
+    remote = "https://github.com/Kindling-project/pixie.git",
+    commit = "2d45f14dc69c060260c5634b7718f05f3f0c7bdb",
 )
 
 BUILD_ALL_CONTENT = """filegroup(name = "all", srcs = glob(["**"]), visibility = ["//visibility:public"])"""
@@ -60,6 +59,10 @@ protobuf_deps()
 load("@com_github_grpc_grpc//bazel:grpc_deps.bzl", "grpc_deps")
 
 grpc_deps()
+
+load("@com_github_jupp0r_prometheus_cpp//bazel:repositories.bzl", "prometheus_cpp_repositories")
+
+prometheus_cpp_repositories()
 
 load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies")
 
--- a/probe/scripts/run_docker.sh
+++ b/probe/scripts/run_docker.sh
@@ -2,14 +2,11 @@ script_dir="$(dirname "$0")"
 workspace_root=$(realpath "${script_dir}/../")
 
 # Docker image information.
-#docker_image_with_tag="kindlingproject/kindling-compiler"
 docker_image_with_tag="kindlingproject/kindling-compiler"
 
 configs=(-v "$HOME/.config:/root/.config" \
   -v "$HOME/.ssh:/root/.ssh" \
-  -v "$HOME/.kube:/root/.kube" \
-  -v "$HOME/.gitconfig:/root/.gitconfig" \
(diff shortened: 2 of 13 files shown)


probe/scripts/run_docker.sh
```diff
@@ -3,13 +3,12 @@ workspace_root=$(realpath "${script_dir}/../")
 
 # Docker image information.
 #docker_image_with_tag="kindlingproject/kindling-compiler"
-docker_image_with_tag="kindlingproject/kindling-compiler"
+# docker_image_with_tag="kindlingproject/kindling-compiler"
+docker_image_with_tag="kindlingproject/kindling-compiler:dev"
```
dxsup, line 7: I think this is unnecessary because once your PR is merged, we will update the image using the tag `latest`.

dxsup (changes requested):

jundizhou:
We should point to this repo:
https://github.com/Kindling-project/pixie

dxsup (approved):

probe/src/probe/publisher/publisher.cpp
```diff
@@ -245,9 +227,8 @@ px::Status publisher::consume_uprobe_data(uint64_t table_id, px::types::TabletID
         auto tinfo = m_inspector->get_thread_ref(pid, true, true, true);
         if (tinfo) {
             gevt.container_id = tinfo->m_container_id;
-            // std::cout << "[qianlu] find container_id for pid:" << pid << " container_id:" << gevt.container_id << std::endl;
         } else {
-            std::cout << "[qianlu] cannot find container_id for pid:" << pid << std::endl;
+		LOG(WARNING) << "[stirling] cannot find container_id for pid:" << pid << std::endl;
```
jundizhou, line 231: I think there will be performance problems here

probe/WORKSPACE
```diff
@@ -4,7 +4,7 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl","git_repository")
 git_repository(
     name = "px",
     remote = "https://github.com/KayzzzZ/pixie.git",
-    commit = "6b853b96987e570e0979620537c23df5850c7b76",
+    commit = "2d45f14dc69c060260c5634b7718f05f3f0c7bdb",
```
jundizhou, line 7: Please change repo to kindling-project/pixie

probe/src/probe/converter/uprobe_converter.cpp
```diff
@@ -48,6 +48,20 @@ void uprobe_converter::convert(void *evt) {
...
+  auto sa_attr = kevt->add_user_attributes();
+  sa_attr->set_key("source_addr");
+  auto sa_value = new AnyValue();
+  sa_value->set_string_value(gevt->source_addr);
+  sa_attr->set_allocated_value(sa_value);
+
+  // source_port
+  auto sp_attr = kevt->add_user_attributes();
+  sp_attr->set_key("source_port");
+  auto sp_value = new AnyValue();
```
jundizhou, line 61: Would it be better to reuse anyvalue?

probe/src/probe/publisher/publisher.cpp
KayzzzZ: You mean this block will be executed frequently?

probe/WORKSPACE
KayzzzZ: ok ...

probe/src/probe/converter/uprobe_converter.cpp
KayzzzZ: Protobuf will manage the object(AnyValue)'s lifestyle. Which means Protobuf will destroy the object but we don't know when and there is no callback... So I think we can't reuse the memory.

probe/src/probe/publisher/publisher.cpp
dxsup: How frequently would it log?

KayzzzZ: It logs when we failed to find the corresponding `container_id` from `sinsp`.

jundizhou: I think it's normal for the virtual machine process to have a request. Is it necessary to print this line of log

jundizhou (approved):

Issue #112, opened 2022-03-15
Print version information when the program starts


dxsup:
**Is your feature request related to a problem? Please describe.**
Whenever an error occurs, we want to know clearly which version of codes the agent is on. But now there is no such information printed by the collector and only an outdated `KINDLING_PROBE_VERSION: v0.1-2021-1221` printed by the probe. We can't know the exact versions of them now. 

**Describe the solution you'd like**
The best version information is the combination of the release version and the commit revision number, IMO. We could insert these numbers as environment variables when we build the containers and print them periodically while the agent is running.

**Additional context**
Any ideas about how we could get this information are appreciated.

zjuzzb:
I find a solution for adding build numbers from git,  take a look at [this](https://pukkaone.github.io/2010/12/19/build-number-git-repository.html).

dxsup:
> I find a solution for adding build numbers from git, take a look at [this](https://pukkaone.github.io/2010/12/19/build-number-git-repository.html).

Looks great! I think it is what we want. We could insert it as a parameter when building the collector, like `go build -ldflags -X main.githash=xxx". Do you have any idea about how we insert it into the probe?

zjuzzb:
> Looks great! I think it is what we want. We could insert it as a parameter when building the collector, like `go build -ldflags -X main.githash=xxx". Do you have any idea about how we insert it into the probe?

```cmake
add_definitions(-DBUILD_NUMBER)
find_package(Git)
if(GIT_FOUND)
    execute_process(
            COMMAND ${GIT_EXECUTABLE} describe --match build
            OUTPUT_VARIABLE BUILD_NUMBER
            OUTPUT_STRIP_TRAILING_WHITESPACE)
    add_definitions(-DBUILD_NUMBER=\ "${BUILD_NUMBER}\")
endif()

```
Add this code into CMakeLists.txt of probe, then just use `BUILD_NUMBER` macro in the source code.

dxsup:
> 
> ```cmake
> add_definitions(-DBUILD_NUMBER)
> find_package(Git)
> if(GIT_FOUND)
>     execute_process(
>             COMMAND ${GIT_EXECUTABLE} describe --match build
>             OUTPUT_VARIABLE BUILD_NUMBER
>             OUTPUT_STRIP_TRAILING_WHITESPACE)
>     add_definitions(-DBUILD_NUMBER=\ "${BUILD_NUMBER}\")
> endif()
> ```
> 
[...]

PR #152, opened 2022-03-25
Print the version of collector when starting

sugary199:
The function of printing commit version for collector is realized.
The version information will be printed before starting applications.

The core.go file is temporarily used in my warehouse. If it can be merged, it's can be changed to `github.com/Kindling-project/kindling/collector/core.CodeVersion`


sugary199:
After I did circular printing, I found that this function is unnecessary and redundant.... so I deleted it

collector/core/core.go
```diff
@@ -0,0 +1,24 @@
+// core.go
+
+package core
```
dxsup, line 3: I don't think it is necessary to introduce a new package considering there is only one file that functions to print versions. Even if we introduce it, I think the package name `core` is kind of vague since this package only contains the functions for printing versions.
Did you refer to any other open-source repositories for introducing a `core` package? I prefer to move this file just under the `collector` directory.

collector/go.mod
```diff
@@ -11,6 +11,7 @@ require (
 	github.com/pebbe/zmq4 v1.2.7
 	github.com/spf13/viper v1.10.1
 	github.com/stretchr/testify v1.7.0
+	github.com/sugary199/collector-version v0.0.0-20220323082224-e39a32dfb492
```
dxsup, line 14: I have checked this repository and found it is identical to `core.go`. This could be removed as you have introduced `core.go` to our base.

collector/main.go
```diff
@@ -3,15 +3,24 @@ package main
 import (
 	"github.com/Kindling-project/kindling/collector/application"
 	"log"
+	"fmt"
+
+	"github.com/sugary199/collector-version/core"
```
dxsup, line 8: Same as the `go.mod` file. I think this line could be replaced with `core/core.go` (before you rename that package).

collector/main.go
```diff
@@ -3,15 +3,24 @@ package main
 import (
 	"github.com/Kindling-project/kindling/collector/application"
 	"log"
+	"fmt"
```
dxsup, line 6: The compiler would complain about an error if an imported package is not used. I don't find the usage of `fmt` in this file so I think it is safe to remove this line.

collector/collector-build.sh
```diff
@@ -0,0 +1,14 @@
...
+# GitCommit=$(git describe)
+GitCommit=$(git rev-parse --short HEAD || echo unsupported)
+echo "Git commit:" $GitCommit
+
+
+# docker run -it -v $PWD:/collector kindlingproject/kindling-collector-builder bash 
+docker run -it -v $PWD:/collector kindlingproject/kindling-collector-builder bash -c 'go build -o kindling-collector -ldflags="-X 'github.com/sugary199/collector-version/core.CodeVersion=$GitCommit'"'
+
+docker build -t kindling-collector -f deploy/Dockerfile .
+
```
dxsup, line 14: This is OK if you just want to contribute for #112. But I prefer to use a Makefile to implement this feature. It will introduce a little complexity but is best practice for a golang project. Anyway, it depends on you. We have had a plan for writing a Makefile later.

dxsup (changes requested):

dxsup:
Thank you for your contribution! But there are a few questions that need your answers. If you need any help, I am always available for access. I will help you polish this PR and merge it into our mainline at the last.

collector/main.go
sugary199: `version` is truly better. I'll change it if we still use this method.

I did want to introduce a package. I see that our package references are imported from the GitHub repository, so I use my warehouse temporarily since  `core` has not been merged into the formal kindling repository.

collector/main.go
sugary199: Oh, sorry! I slightly modified it before submitting but did not test it. I was negligent. Speaking of this, I would like to ask whether any changes to the program should be tested through the whole process of compiling, mirroring, uploading and deploying to the cluster. For example, this function just does small changes in the `main()`, but it takes a long time to verify it through the whole process.

sugary199:
> Thank you for your contribution! But there are a few questions that need your answers. If you need any help, I am always available for access. I will help you polish this PR and merge it into our mainline at the last.

Thank you so much for your generous help! To complete this function, do I need to learn how to use makefile and use the Makefile method instead?

collector/main.go
dxsup: Sorry for my incorrect suggestion. `github.com/sugary199/collector-version/core` could be replaced with `github.com/Kindling-project/kindling/collector/core` instead. You can see our module name in the first line of `go.mod` is `github.com/Kindling-project/kindling/collector` which means you could just import the packages that belong to the module with its module name beginning.

dxsup:
Since you asked this question, I think you better not do that in this PR. If you are interested in how to build a Go project using Makefile, you can open another issue and relate a new PR with it, after which you can learn how to use Makefile and work on that feature.

collector/main.go
sugary199: 1、I‘ve replaced 'core' with 'version'.
2、Comments are added in front of the commands and codes that need to be modified if they are merged.
The specific location is 
'//  "github.com/Kindling-project/kindling/collector/version"' in 'main.go'
'# docker run -it -v $PWD:/collector kindlingproject/kindling-collector-builder bash -c 'go build -o kindling-collector -ldflags="-X 'github.com/Kindling-project/kindling/collector/version.CodeVersion=$GitCommit'"'
' in 'collector-build.sh'

Is this right?

sugary199: 1、I‘ve replaced `core` with `version`.
2、Comments are added in front of the commands and codes that need to be modified if they are merged.
The specific location is
`// "github.com/Kindling-project/kindling/collector/version"` in `main.go`
`# docker run -it -v $PWD:/collector kindlingproject/kindling-collector-builder bash -c 'go build -o kindling-collector -ldflags="-X 'github.com/Kindling-project/kindling/collector/version.CodeVersion=$GitCommit'"`
in `collector-build.sh`

Is this right?

collector/main.go
```diff
@@ -3,15 +3,22 @@ package main
...
 )
 
 func main() {
 	app, err := application.New()
 	if err != nil {
 		log.Fatalf("Failed to create application: %v", err)
 	}
+	
+	//print version information
+	log.Printf("GitCommitInfo:%s\n", version.Version())
```
dxsup, line 16: I would rather move this to the beginning of the main function in which way even if users can not create `application`, we could also know the version.

collector/collector-build.sh
```diff
@@ -0,0 +1,13 @@
+#collector
+#cd kindling/collector
+
+# If developers start labeling later:
+# GitCommit=$(git describe)
+GitCommit=$(git rev-parse --short HEAD || echo unsupported)
+echo "Git commit:" $GitCommit
+
+
+docker run -it -v $PWD:/collector kindlingproject/kindling-collector-builder bash -c 'go build -o kindling-collector -ldflags="-X 'github.com/Kindling-project/kindling/collector/version.CodeVersion=$GitCommit'"'
```
dxsup, line 10: I think the file `collector-build.sh` should focus only on "building" and "injecting" git revision number. Anything about `docker` should move out of this file, in which way we could combine freely multiple scripts later. I think this is also one of the practices of the single responsibility principle.

collector/main.go
```diff
@@ -3,15 +3,22 @@ package main
 import (
 	"github.com/Kindling-project/kindling/collector/application"
 	"log"
+    "github.com/Kindling-project/kindling/collector/version"
```
dxsup, line 6: Please use `gofmt` tool to format these lines.

sugary199: Correction： 
1、Standardized the format of `main.go` and print at the beginning   
2、Changed the mount point of the collector-build container
3、Separated startup and compilation. However, because the change of the mount point changes the container startup command, I don't know where to put this command temporarily, so I left this startup command in  `collector-version-build.sh`.

dxsup (approved):

PR #153, opened 2022-03-28
replace proto by gogoproto to improve performance

sanyangji:
Signed-off-by: sanyangji <songyujie@zju.edu.cn>


dxsup:
There is a little improvement after moving to gogo-protobuf.

dxsup (approved):

PR #156, opened 2022-03-29
add send log

sanyangji:
One log includes the number of current and total sending kindling events, the size of kindling events after marshaling.
Use GLOG.

### DIFF
--- a/probe/src/probe/main.cpp
+++ b/probe/src/probe/main.cpp
@@ -76,30 +76,31 @@ void do_inspect(sinsp *inspector, sinsp_evt_formatter *formatter, int pid, publi
 }
 
 void get_capture_statistics(sinsp* inspector) {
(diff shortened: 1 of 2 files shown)


jundizhou:
I'm working on remove pixie.Therefore, these changes are not necessary for the time being

sanyangji:
I think the send log should be necessary. Although GLOG is imported by pixie, we can import directly in the future. Or we can use other log framework, but for now, using GLOG is a better choice. @jundizhou

dxsup:
I think it is long-term work for removing Pixie. What this PR intends to contribute is that improve the observability of the probe by printing logs of how many events are sent per second. I think this feature is necessary. The only problem about pixie is whether we should depend on GLOG which is provided by Pixie.

jundizhou (approved):

PR #158, opened 2022-03-29
fix topology plugin 

thousandxu:
1.Switch the topology layout from the configuration to the page
2.Fix type error when pod name does not exist
3.fix legned does not support node error
4.add package lost metric
5.fix error rate calculate error
6.Modify the plugin configuration to conform to component open source standards, use kindlingproject rename topology plugin ID
7.modify dashboard promethues query, use $_range instead of $_interval


grafana-plugins/topo-plugin/src/topology/services.ts
```diff
@@ -234,6 +234,39 @@ export const nsRelationHandle = (topoData: any, nodeData: NodeDataProps, edgeDat
...
+                source: source,
+                target: target,
+                service: service || '',
+                opposite
+            });
+        }
+    });
+    nodes = detailNodesHandle(nodes, nodeData);
+    edges = detailEdgesHandle(nodes, edges, edgeData, serviceLine);
+    return { nodes, edges };
```
jundizhou, line 268: Please use English notes

dxsup (approved):

PR #159, opened 2022-03-30
WIP: Pref/use a adapter to replace format processor to avoid recreate gauge during relabling

NeJan2020:
In the performance tests, we can see most memory malloc requests happened in format processor and otlp-exporter, which lead to a huge cost of GC. 
This PR is an implementation of a metric adapter in the otlp-exporter, which can use a static paramMap to relabel our gauges and pass this to the exporter.

### DIFF
--- a/collector/consumer/exporter/otelexporter/adapter.go
+++ b/collector/consumer/exporter/otelexporter/adapter.go
@@ -7,153 +7,258 @@ import (
 	"go.opentelemetry.io/otel/metric"
 )
 
-type valueType int
-
-type dictionary struct {
-	originKey string
-	newKey    string
-	valueType
-}
-
-const (
-	Int64  valueType = 0
-	String valueType = 1
-	Bool   valueType = 3
-)
-
-var entityCommonMetricsDicList = []dictionary{
-	{"", "", String},
-}
-
-var entityAggMetricsDicList = []dictionary{
-	{"", "", String},
-}
-
-var entityDetailMetricsDicList = []dictionary{
-	{"", "", String},
-}
-
-var topologyCommonMetricsDicList = []dictionary{
-	{"", "", String},
-}
-
-var topologyAggMetricsDicList = []dictionary{
-	{"", "", String},
-}
-
-var topologyDetailMetricsDicList = []dictionary{
-	{"", "", String},
-}
-
-var ServerEntity *metricAdapter = creatMetricAdapter(nil, entityCommonMetricsDicList, true, nil)
-
 type metricAdapter struct {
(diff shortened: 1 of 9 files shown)


NeJan2020:
# Design of adapter

## Change

Use a new internal component named adapter to replace the format processor we used to relabel output metrics.

## Why we made this change

In performance testing, we can see that collector costs lots of CPU to generate attributes for Gauges. Those attributes is short-lived and also increase the cost of GC.

The main purpose to use the adapter is trying to use a static memory space for general attributes we needed, pass this to otlp-SDK, and reuse this space after we record this space.

The format processor has the same problem just like the exporter. But it's much more complicated to reuse this space since the general gauges will pass to other consumers and may be modified.

So we use an internal component in the exporter named adapter, which can replace format processor so that all memory costs in relabeling and generating attributes can be cut down to const.

## Struct

The generated attributes can be apart into three parts: topologyAndInstanceLabels, protocol label, and const labels.

The topologyAndInstanceLabels contains IP, port, and k8s info of source and destination. In kindling metrics, there is four kinds of topologyAndInstanceLabels: DetailTopology, DetailEntity, AggTopology, AggEntity.

The protocol labels are determined by the protocol type of the original data. Each protocol has its own labels.

For topologyAndInstanceLabels, we create four adapters to match each kind of kindlingMetrics.

For protocol label, we use a map to store different labels and use the protocol as the key of the map.

Here is the struct of the adapter.

``` go
type Adapter struct {
	labelsMap  map[protocol]realAttributes
	updateKey func

    //...
}

type realAttributes struct {
	// paramMap A list contain baseLabels,commonLabels,extraLabelsParamList,valueLabels,constLabels
	paramMap []attribute.KeyValue
	// metricsDicList A list contain dict of baseLabels,commonLabels,extraLabelsParamList,
	metricsDicList []dictionary

	// ...
}

type dictionary struct {
	newKey    string
	originKey string
	valueType
}
```

In Adapter, we use the function named `updateKey` to generate the key of `labelsMap`(just protocol now), for each protocol, we can get a `realAttributes` which contains the `paramMap` and `metricsDicList`.

The `metricsDicList` is a list of dictionaries, which explain how we generalize the attributes from original data.

The `paramMap` is the static obj which we will reuse, which has the same order as paramMap, but at the end of paramMap, it appended the const labels which is not mentioned in the `metricsDicList`.

So we can traverse the `metricsDicList`, using the `originKey` to get info from original data, using the `valyeType` to decide how to convert, using the `newKey` or index of `originKey` to modify the value in the paramMap. After traversing, the adapter uses the paramMap as the final output to otlp-SDK.

PR #160, opened 2022-03-30
[WIP] optimize the structure of kindling event

sanyangji:
### Motivation
We notice that the current struct of kindling event is sort of bad for performance so we have to make some change on it to improve performance.
The first step is to replace AnyValue with KeyValue that consists of name, valueType and the value in format of bytes. Using AnyValue, we can easily to get value while it adds pressure for object allocation and GC. Instead, we can infer the attributes based on KeyValue which may save more resources (CPU and MEM).
### Benchmark
I make a benchmark on the change, Postfix `AnyValue` indicates before changing, and Postfix `Opt`indicates after changing. Each event contains the same event information and the only difference is using AnyValue or KeyValue to store user_attributes. 
`BenchmarkMarshal` is the test that marshal 5, 000 kindling events;
`BenchmarkUnMarshal` is the test that unmarshal 5, 000 kindling events;
`BenchmarkUnMarshalAndRange` is the test that range 5, 000 kindling events and get the UserAttributes aftering unmarshaling.
```
BenchmarkMarshalAnyValue
BenchmarkMarshalAnyValue-4             	      74	  13686098 ns/op
BenchmarkMarshalOpt
BenchmarkMarshalOpt-4                  	     230	   5049329 ns/op
BenchmarkUnMarshalAnyValue
BenchmarkUnMarshalAnyValue-4           	      85	  12575498 ns/op
BenchmarkUnMarshalOpt
BenchmarkUnMarshalOpt-4                	     100	  10728335 ns/op
BenchmarkUnMarshalAndRangeAnyValue
BenchmarkUnMarshalAndRangeAnyValue-4   	      86	  13804747 ns/op
BenchmarkUnMarshalAndRangeOpt
BenchmarkUnMarshalAndRangeOpt-4        	     130	   8595758 ns/op
```
The result shows that the change on the structure of AnyValue does improve the performance.
#### Test on Probe
Before Changing
image
After Changing
image
The Marshal part behaves similarly while the clear of kindling events behaves better. The destructor of kindling events cost less. (decrease from 12.64% to 8.67%)
#### Test on Collector
The TPS is about 10-11K.
Before Changing (remove exporter)
image
After Changing (remove exporter)
image
We can see the improvement on Unmarshal part, which decrease from 32.38% to 25.40%. The improvement of GC is not obvious.


collector/model/kindling_event_helper.go
```diff
@@ -48,14 +59,56 @@ func (x *KindlingEvent) GetLatency() uint64 {
 	return 0
 }
 
-func (x *KindlingEvent) GetUint64UserAtrribute(key string) uint64 {
+func (x *KindlingEvent) GetUintUserAtrribute(key string) uint64 {
```
sanyangji, line 62: All get method may have typo problem, should be Attribute.

collector/analyzer/tcpmetricanalyzer/tcp_analyzer.go
```diff
@@ -144,10 +145,10 @@ func (a *TcpMetricAnalyzer) getTupleLabels(event *model.KindlingEvent) (*model.A
...
 		return nil, fmt.Errorf("one of sip or dip or dport is nil for event %s", event.Name)
 	}
-	sIpString := model.IPLong2String(uint32(sIp.Value.GetUintValue()))
-	sPortUint := sPort.Value.GetUintValue()
-	dIpString := model.IPLong2String(uint32(dIp.Value.GetUintValue()))
-	dPortUint := dPort.Value.GetUintValue()
+	sIpString := model.IPLong2String(uint32(event.GetUintUserAttribute("sip")))
+	sPortUint := event.GetUintUserAttribute("sport")
+	dIpString := model.IPLong2String(uint32(event.GetUintUserAttribute("dip")))
+	dPortUint := event.GetUintUserAttribute("dport")
```
dxsup, line 151: There are several variables above called `sIp`, `sPort`, `dIp`, and `dPort`  which could be reused. I think we better check the value before using it as there are a lot of empty values for these events.

dxsup (changes requested):

dxsup (approved):
LGTM

PR #163, opened 2022-03-31
Create pull_request_template.md

dxsup:

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -5,7 +5,7 @@
 
 ## Related Issue
 <!--- If suggesting a new feature or change, please discuss it in an issue first -->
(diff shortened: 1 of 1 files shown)


J-lena (changes requested):
If fixing a bug, there should be an issue describing it with steps to reproduce. i think it seems needless.

 "If fixing an existing bug , please link to the issue here "  and "If fixing an unpublished bug, please post this bug as an issue " will be better

dxsup:
To be honest, I don't see much difference. What we want contributors to do is that link the PR with an issue. I have modified some words to express it more clearly.
```
## Related Issue
<!--- If suggesting a new feature or change, please discuss it in an issue first -->
<!--- If fixing a bug, there should be an issue describing it with details -->
<!--- Please link to the issue here: -->
```

PR #165, opened 2022-03-31
Fix MultiTopics in kafka payload will be discerned as NOSUPPORT

hocktea214:
## Description
The kafka data is discerned as kafka and NOSUPPORT

## Motivation and Context
There is an strict requirements for only one topic with Kafka payload, but in fact there may be more than one.
When there is multi topics payload, the protocol is discerned as NOSUPPORT.

* When there is more than one topic, the topicName will take the first name based on following case:
    * The payload is substr with fixed length(1K), it's not able to get all topic names
    * If we get enough length for payload(Maybe 10K), the parser will take more performance cost
    * There is not enough cases to cover multi-topics
* The partition is not used, just remove it.

## How Has This Been Tested?
Run the unit test under `collector/analyzer/network` with `TestKafkaProtocol()` and the result is as follows.
```
=== RUN   TestKafkaProtocol
=== RUN   TestKafkaProtocol/produce-split
=== RUN   TestKafkaProtocol/fetch-split
=== RUN   TestKafkaProtocol/fetch-one-topic-two-patitions
--- PASS: TestKafkaProtocol (0.00s)
    --- PASS: TestKafkaProtocol/produce-split (0.00s)
    --- PASS: TestKafkaProtocol/fetch-split (0.00s)
    --- PASS: TestKafkaProtocol/fetch-one-topic-two-patitions (0.00s)
PASS
```

### DIFF
--- a/collector/analyzer/network/protocol/kafka/kafka_request_fetch.go
+++ b/collector/analyzer/network/protocol/kafka/kafka_request_fetch.go
@@ -14,12 +14,10 @@ func fastfailRequestFetch() protocol.FastFailFn {
 func parseRequestFetch() protocol.ParsePkgFn {
 	return func(message *protocol.PayloadMessage) (bool, bool) {
 		var (
-			offset       int
-			err          error
-			topicNum     int32
-			topicName    string
-			partitionNum int32
-			partition    int32
+			offset    int
+			err       error
+			topicNum  int32
+			topicName string
 		)
 		version := message.GetIntAttribute(constlabels.KafkaVersion)
 		compact := version >= 12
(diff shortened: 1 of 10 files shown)


collector/analyzer/network/protocol/kafka/kafka_request_fetch.go
```diff
@@ -36,20 +34,16 @@ func parseRequestFetch() protocol.ParsePkgFn {
...
-		if offset, err = message.ReadArraySize(offset, compact, &partitionNum); err != nil || partitionNum != 1 {
-			return false, true
-		}
-		if _, err = message.ReadInt32(offset, &partition); err != nil {
-			return false, true
+		if topicNum > 0 {
+			if _, err = message.ReadString(offset, compact, &topicName); err != nil {
+				return false, true
+			}
+			// Read First TopicName
```
dxsup, line 44: Could you explain why this is done in comments?

hocktea214: Based on following case, we	just read first topicName.

1. The payload is substr with fixed length(1K), it's not able to get all topic names
2. If we get enough length for payload(Maybe 10K), the parser will take more performance cost
3. There is not enough cases to cover multi-topics

dxsup (approved):

PR #170, opened 2022-04-02
Add source ip and port for the uprobe events

KayzzzZ:
- Update Pixie module, support obtain source addr & port.
- Add log when uprobe_converter list is full.
Signed-off-by: qianlu.kk [qianlu.lk@alibaba-inc.com](mailto:qianlu.lk@alibaba-inc.com)

### DIFF
--- a/probe/WORKSPACE
+++ b/probe/WORKSPACE
@@ -4,7 +4,7 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl","git_repository")
 git_repository(
     name = "px",
     remote = "https://github.com/Kindling-project/pixie.git",
-    commit = "2d45f14dc69c060260c5634b7718f05f3f0c7bdb",
+    commit = "c98c1e8c041038d535db1fb2590087dcb91cf92f",
 )
 
 BUILD_ALL_CONTENT = """filegroup(name = "all", srcs = glob(["**"]), visibility = ["//visibility:public"])"""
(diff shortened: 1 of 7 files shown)


sanyangji (approved):

probe/WORKSPACE
```diff
@@ -4,7 +4,7 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl","git_repository")
 git_repository(
     name = "px",
     remote = "https://github.com/Kindling-project/pixie.git",
-    commit = "2d45f14dc69c060260c5634b7718f05f3f0c7bdb",
+    commit = "c98c1e8c041038d535db1fb2590087dcb91cf92f",
```
dxsup, line 7: I found this refers to a branch called `kindling-20220402`. How should we manage the branches of that repository? I don't think it's a good idea to create a new branch every time there is a change. Could you merge that branch to a continuous developing one, like `kindling-dev`, if you have verified the feature works as expected? I think we better track all changes in one branch.

KayzzzZ: I think our kindling-dev branch needs to be a relatively stable version. However, not every change can be called "stable". Therefore, I may change the branch name every time I commit. 
As for the branch management stuff, my solution is that I will clean up some deprecated branches, and I will periodically rebase stable releases to `kindling-dev` branch.
As a result, you can find the latest beta branch through the commit id, and if there is some problems with this branch, you can also manually roll back to the stable version of `kindling-dev`.

dxsup (approved):

Issue #179, opened 2022-04-15
Improve aggregation performance of the collector 


dxsup:
**Is your feature request related to a problem? Please describe.**
Currently, the collector has a performance issue on metrics aggregation. 
![untitle](https://user-images.githubusercontent.com/46807570/163513621-fc67d088-6c17-4ebc-a542-c8c0efd576f8.png)

**Describe the solution you'd like**
Write an `aggregateprocessor` to aggregate the `GaugeGroup` before the `otelexporter`.

**Describe alternatives you've considered**
Profile the opentelemetry-go SDK to improve its performance.

dxsup:
Closed by #180.

Issue #177, opened 2022-04-15
Improve metrics aggregation and redesign the metric model


dxsup:
## Background
After analyzing the flame graph of the collector, we found that the performance of the opentelemetry-go SDK could not meet our requirements, so we implemented an aggregation module by ourselves. According to the benchmark, the new module at least doubles the performance of collectors.

However, to use this aggregation module, it is necessary to make changes to the existing metrics data model. The main cause is that the original `histogram` isn't functional anymore.

### Problem 1: Changes in the histogram's meaning
The original histogram utilized the opentelemetry-go SDK to count the "duration of each request", and finally generated three sub-metrics, including "total number of requests", "total request duration", and "buckets of request duration". We can use PromQL to calculate the quantiles based on the bucket metrics.

However, after we introduced the aggregation module, the "duration of each request" was aggregated in advance, and it was impossible to transfer the exact single duration to the SDK. What we could transfer is the average duration. Then three sub-metrics were generated, including "total number of average duration", "sum of average durations" and "buckets of average duration", from which we can calculate the quantile of "average delay" based on the bucket.

The results calculated by these two different methods are both approximate values, but due to counting more numbers, a more accurate value will be obtained by the first method.

### Problem 2: There are two metrics missing, "Total Requests" and "Total Request Duration"
The second problem brought about by the introduction of the aggregation module is that the two metrics, "Total Requests" and "Total Request Duration", are missing and need to be added manually.

### Problem 3: Lack of hidden metric, "Number of Slow Requests"
Another hidden metric that we had before was "Number of Slow Requests", which is calculated by counting the number of bucket metrics that is greater than a certain boundary. However, since the bucket metrics now count the number of "average duration", the original method is invalid. We have to add it manually.

## New Data Model
Based on the above consideration, we better redesign the data model.
### Transformation
- kindling_entity_request_duration_nanoseconds_bucket -> kindling_entity_request_average_duration_nanoseconds_bucket
- kindling_entity_request_duration_nanoseconds_sum -> kindling_entity_request_duration_nanoseconds_total
- kindling_entity_request_duration_nanoseconds_count -> kindling_entity_request_total

- kindling_topology_request_duration_nanoseconds_bucket 
-> kindling_topology_request_average_duration_nanoseconds_bucket
- kindling_topology_request_duration_nanoseconds_sum -> kindling_topology_request_duration_nanoseconds_total
- kindling_topology_request_duration_nanoseconds_count -> kindling_topology_request_total
### New Label
[...]

PR #180, opened 2022-04-15
Add an aggregateprocessor to aggregate gauge before the exporter

dxsup:
## Description
Add an `aggregateprocessor` to aggregate GaugeGroup before the exporter.

## Related Issue

#179 and #177 

## Motivation and Context

This will improve the performance of the collector greatly. The following is the benchmark result.
#### TPS is about 5400.
Pasted Graphic 4
#### The average duration is 1.9ms
Pasted Graphic 5
#### The P95 is about 9.5ms
Pasted Graphic 6
#### The CPU usage is about 0.95,
Pasted Graphic 7

#### The TPS is about 6000.
Pasted Graphic 1
#### The average duration is 1.8ms.
Pasted Graphic 2
#### The P95 is 9.5ms.
Pasted Graphic 3
#### The CPU usage is 0.33.
Pasted Graphic

## How Has This Been Tested?

All unit tests are run and the metrics are tested under developing environment.


collector/application/application.go
```diff
@@ -106,17 +108,20 @@ func (a *Application) buildPipeline() error {
...
+	k8sMetadataProcessor := k8sProcessorFactory.NewFunc(k8sProcessorFactory.Config, a.telemetry.Telemetry, aggregateProcessor)
 	// Initialize all analyzers
 	// 1. Common network request analyzer
 	networkAnalyzerFactory := a.componentsFactory.Analyzers[network.Network.String()]
 	// Now NetworkAnalyzer must be initialized before any other analyzers, because it will
 	// use its configuration to initialize the conntracker module which is also used by others.
 	networkAnalyzer := networkAnalyzerFactory.NewFunc(networkAnalyzerFactory.Config, a.telemetry.Telemetry, []consumer.Consumer{k8sMetadataProcessor})
 	// 2. Layer 4 TCP events analyzer
-	k8sMetadataProcessor2 := k8sProcessorFactory.NewFunc(k8sProcessorFactory.Config, a.telemetry.Telemetry, otelExporter)
+	aggregateProcessorForTcp := aggregateProcessorFactory.NewFunc(aggregateProcessorFactory.Config, a.telemetry.Telemetry, otelExporter)
```
NeJan2020, line 123: Due to different subsequent processing components, multiple instances need to be created; The same problem will also occur on conntrack (if we later treat it as a separate processor). I think this problem can be avoided if the information of subsequent nodes of the component is maintained by an external manager

dxsup: I'm not sure whether I got your point correctly. I think multiple instances won't bring many problems. Why do we need an external manager?

NeJan2020 (approved):

collector/deploy/kindling-collector-config.yml
```diff
@@ -79,14 +79,40 @@ processors:
...
     need_trace_as_span: false
+  aggregateprocessor:
+    # Aggregation duration window size. The unit is second.
+    ticker_interval: 5
+    aggregate_kind_map:
+      request_total_time:
+        - kind: sum
+        - kind: avg
+          output_name: request_total_time_avg
+        - kind: count
```
NeJan2020, line 90: we also need a `last` value for request_total_time for trace as metric

collector/consumer/processor/aggregateprocessor/internal/label_key.go
```diff
@@ -0,0 +1,73 @@
...
+		case BooleanType:
+			value, _ := strconv.ParseBool(key.Value)
+			ret.AddBoolValue(key.Name, value)
+		case StringType:
+			ret.AddStringValue(key.Name, key.Value)
+		case IntType:
+			value, _ := strconv.ParseInt(key.Value, 10, 64)
+			ret.AddIntValue(key.Name, value)
+		}
+	}
```
NeJan2020, line 71: I think using the `for i` instead of `for range` can get better performance here

NeJan2020 (dismissed):
LGTM, just add some preformance advice

collector/application/application.go
dxsup: Processors must be able to be created multiple times as we will make it possible that the pipeline could be configured later. For those components that don't support multiple instances, refactoring will be conducted.

collector/consumer/processor/aggregateprocessor/internal/label_key.go
dxsup: Actually, they perform nearly identically.

collector/deploy/kindling-collector-config.yml
dxsup: There is no `traceAsmetric` generated when the `aggregateprocessor` consumes data. I'd like to work around this problem later.

NeJan2020 (dismissed):

NeJan2020 (approved):

NeJan2020 (approved):

collector/consumer/processor/aggregateprocessor/internal/label_key.go
dxsup: Right, here `k.keys` is an array. Iterating an array using `for range value` will copy the original value. But since this method is called less frequently, the impact on performance is much little.

PR #185, opened 2022-04-19
Add Prometheus metrics description doc

### LINKED ISSUE #177: Improve metrics aggregation and redesign the metric model

dxsup:
## Description

Add Prometheus metrics description doc.

## Related Issue

This PR will close #177 
## Motivation and Context

See #177

### DIFF
--- a/docs/prometheus_metrics.md
+++ b/docs/prometheus_metrics.md
@@ -1,0 +1,198 @@
+# Prometheus Metrics Description
+## Service Metrics
+Service metrics are generated from the server-side events, which are used to show the quality of service. 
+| **Metric Name** | **Type** | **Description** |
+| --- | --- | --- |
+| kindling_entity_request_total | Counter | Total number of requests |
+| kindling_entity_request_duration_nanoseconds_total | Counter | Total duration of requests |
+| kindling_entity_request_send_bytes_total | Counter | Total size of payload sent |
+| kindling_entity_request_receive_bytes_total | Counter | Total size of payload received |
+| kindling_entity_request_average_duration_nanoseconds_count  | Counter(Histogram) | Count of average duration of requests |
+| kindling_entity_request_average_duration_nanoseconds_sum | Counter(Histogram) | Sum of average duration of requests |
+| kindling_entity_request_average_duration_nanoseconds_bucket | Counter(Histogram) | Histogram buckets of average duration of requests |
+
+| **Label Name** | **Example** | **Notes** |
+| --- | --- | --- |
+| node | worker-1 | Node name represented in Kubernetes cluster |
+| namespace | default | Namespace of the Pod |
+| workload_kind | daemonset | K8sResourceType |
+| workload_name | api-ds | K8sResourceName |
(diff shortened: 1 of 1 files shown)


J-lena (approved):

NeJan2020 (dismissed):

NeJan2020 (approved):
LGTM

PR #186, opened 2022-04-19
refactor/use adapter to replace format-processor

NeJan2020:
In the performance tests, we can see most memory malloc requests happened in format-processor and otlp-exporter, which lead to a huge cost of GC.
This PR is an implementation of a metric adapter in the otlp-exporter, which can use a static paramMap to relabel our gauges and pass this to the exporter.

This PR has also changed an existed impl of recordGaugeAsync mentioned in #159.

## Description
Use a new internal component named adapter to replace the format-processor we used to relabel output metrics.

## Motivation and Context
In performance testing, we can see that collector costs lots of CPU to generate attributes for Gauges. Those attributes is short-lived and also increase the cost of GC.

The main purpose to use the adapter is trying to use a static memory space for general attributes we needed, pass this to otlp-SDK, and reuse this space after we record this space.

The format processor has the same problem just like the exporter. But it's much more complicated to reuse this space since the general gauges will pass to other consumers and may be modified.

So we use an internal component in the exporter named adapter, which can replace the format-processor so that all memory costs in relabeling and generating attributes can be cut down to const.

## How Has This Been Tested?
All unit tests are run and the metrics are tested in a developing environment.


collector/model/metric.go
```diff
@@ -82,7 +83,7 @@ func (g *GaugeGroup) String() string {
 }
 
 func (g *GaugeGroup) Reset() {
-	g.Name = ""
+	g.Name = constlabels.NetWorkAnalyzeGaugeGroup
```
dxsup, line 86: Why is it reset to this name? The Reset() method should not assume any names.

NeJan2020: Already changed

dxsup:
Since we have introduced an `aggregatorprocessor` put in front of the `kindlingformatprocessor`, the GC impact on the performance is much less now.  I think the main motivation of this PR is to improve the module structure instead. The `processor` should only PROCESS `Gauge Group`. What the `kindlingformatprocessor` does is generate a new `Gauge Group` according to the designed final format, which is actually supposed to be done by the `exporter`.

collector/model/metric.go
```diff
@@ -2,6 +2,7 @@ package model
 
 import (
 	"fmt"
+	"github.com/Kindling-project/kindling/collector/model/constlabels"
```
dxsup, line 5: Error this line.

dxsup (changes requested):

collector/model/metric.go
NeJan2020: done

collector/consumer/exporter/otelexporter/defaultadapter/net_adapter.go
```diff
@@ -0,0 +1,264 @@
...
+	if gaugeGroup.Name != constnames.AggregatedNetRequestGaugeGroup && gaugeGroup.Name != constnames.SingleNetRequestGaugeGroup {
+		// skip non net gaugeGroup
+		return nil, nil
+	}
+
+	if gaugeGroup.Name == constnames.AggregatedNetRequestGaugeGroup {
+		return n.dealWithPreAggGaugeGroups(gaugeGroup)
+	} else {
+		return n.dealWithSingleGaugeGroup(gaugeGroup)
+	}
```
dxsup, line 33: Could this be simplified using `switch-case`?

NeJan2020: Good idea

collector/consumer/exporter/otelexporter/defaultadapter/adapter.go
```diff
@@ -0,0 +1,37 @@
...
+import (
+	"github.com/Kindling-project/kindling/collector/model"
+	"go.opentelemetry.io/otel/attribute"
+)
+
+type Adapter interface {
+	Adapt(group *model.GaugeGroup) ([]*AdaptedResult, error)
+
+	// Transform Only Aggregator Value should use the Transform
+	Transform(group *model.GaugeGroup) (*model.AttributeMap, error)
```
dxsup, line 12: What difference is it between `Adapt` and `Transform`? Could you describe it as comments?

collector/consumer/exporter/otelexporter/defaultadapter/net_adapter.go
```diff
@@ -0,0 +1,262 @@
...
+	for i := 0; i < len(gaugeGroup.Values); i++ {
+		if gaugeGroup.Values[i].Name == constvalues.RequestTotalTime {
+			requestTotalTimeIndex = i
+			break
+		}
+	}
+	if requestTotalTimeIndex == -1 {
+		// No requestTotalTimeGauge
+		return nil, nil
+	}
```
dxsup, line 50: Why not use `GaugeGroup.GetGauge(name string)` directly?

collector/consumer/exporter/otelexporter/consume.go
```diff
@@ -0,0 +1,110 @@
...
+		gauge := result.Gauges[s]
+		var metricName string
+		switch result.RenameRule {
+		case defaultadapter.ServerMetrics:
+			metricName = constnames.ToKindlingNetMetricName(gauge.Name, true)
+		case defaultadapter.TopologyMetrics:
+			metricName = constnames.ToKindlingNetMetricName(gauge.Name, false)
+		default:
+			metricName = gauge.Name
+		}
```
dxsup, line 85: I think this is the responsibility of adapters. Think of if someone wants to add a new adapter, he won't care about `ServerMetrics` or `TopologyMetrics`. What the method `exportMetric` should focus on is only EXPORT metrics. In addition, the parameter `adapter defaultadapter.Adapter` should not be involved.

NeJan2020:
In this PR, We add an interface named 'Adapter', which contains two methods: Adapt and Transform.

Adapt is a method used to deal with all the gaugeGroups received by the otel-exporter. So it should contain the functions below:
1. determined whether this adapter would deal with the incoming gauge groups
2. tell the exporter how to export the data in the struct of AdapterResult

Transform is much different from Adapter, it only deals with those metrics which need to be pre-aggregated. Those metrics were marked as `MAGauge` in config. When exporting those metrics, the exporter will use this adapter to relabel those GaugeGroups and send them into the aggregator.

collector/consumer/exporter/otelexporter/instrument.go
```diff
@@ -58,47 +71,140 @@ func (i *instrumentFactory) createNewInstrument(metricName string, kind MetricAg
...
 }
 
-func (i *instrumentFactory) recordGaugeAsync(metricName string, singleGauge model.GaugeGroup) {
-	i.gaugeChan.put(&singleGauge, i.logger)
-	if i.isGaugeAsyncInitialized(metricName) {
-		return
+// only support TraceAsMetric and TcpRttMills now
+func (i *instrumentFactory) recordLastValue(metricName string, singleGauge *model.GaugeGroup) error {
+	if item, ok := i.aggregators.Load(metricName); ok {
+		aggregator := item.(*defaultaggregator.DefaultAggregator)
```
dxsup, line 77: The `DefaultAggregator` is not designed for use with multiple instances. The best practice is to insert different metrics into a single aggregator to optimize the performance. I know you want to dump different metrics using different observer callbacks, then you better redesign a new aggregator or modify the current implementation.

collector/consumer/exporter/otelexporter/consume.go
```diff
@@ -0,0 +1,96 @@
...
+
+func (e *OtelExporter) exportTrace(result *defaultadapter.AdaptedResult) {
+	if e.defaultTracer != nil && e.cfg.AdapterConfig.NeedTraceAsResourceSpan {
+		_, span := e.defaultTracer.Start(
+			context.Background(),
+			constvalues.SpanInfo,
+			apitrace.WithAttributes(result.Attrs...),
+		)
+		span.End()
+	} else if e.defaultTracer != nil && e.cfg.AdapterConfig.NeedTraceAsResourceSpan {
```
dxsup, line 66: The bool expressions of `if` and `else if` are identical. Any typos here? I think a guard clause is better, that is if `e.defaultTracer` is nil, the function 
 return directly.

collector/consumer/exporter/otelexporter/consume.go
```diff
@@ -0,0 +1,96 @@
...
+				Timestamp: result.Timestamp,
+			})
+			if err != nil {
+				e.telemetry.Logger.Error("Failed to record Gauge", zap.Error(err))
+			}
+		} else if ok {
+			measurements = append(measurements, e.instrumentFactory.getInstrument(metricName, metricKind).Measurement(gauge.Value))
+		} else {
+			measurements = append(measurements, e.instrumentFactory.getInstrument(metricName, MACounterKind).Measurement(gauge.Value))
+		}
```
dxsup, line 91: This branch will change the default behavior of `otelexporter` when the `metricName` is not configured, that was printing logs and skipping the metric.

collector/consumer/exporter/otelexporter/defaultadapter/otel_adapter_cache.go
```diff
@@ -0,0 +1,349 @@
...
+import (
+	"github.com/Kindling-project/kindling/collector/model"
+	"github.com/Kindling-project/kindling/collector/model/constlabels"
+	"github.com/Kindling-project/kindling/collector/model/constvalues"
+	"go.opentelemetry.io/otel/attribute"
+	"sort"
+	"strconv"
+)
+
+type adapterCache struct {
```
dxsup, line 12: Could you add a comment on this line to describe what `adapterCache` does and why it is named as "cache"?

collector/consumer/exporter/otelexporter/consume.go
```diff
@@ -0,0 +1,96 @@
...
+			constvalues.SpanInfo,
+			apitrace.WithAttributes(result.Attrs...),
+		)
+		span.End()
+	} else if e.defaultTracer != nil && e.cfg.AdapterConfig.NeedTraceAsResourceSpan {
+		e.telemetry.Logger.Error("send span failed: this exporter can not support Span Data", zap.String("exporter", e.cfg.ExportKind))
+	}
+}
+
+func (e *OtelExporter) exportMetric(result *defaultadapter.AdaptedResult, adapter defaultadapter.Adapter) {
```
dxsup, line 71: The parameter `adapter` is no longer necessary.

collector/consumer/exporter/otelexporter/consume.go
```diff
@@ -0,0 +1,97 @@
...
+
+func (e *OtelExporter) exportTrace(result *defaultadapter.AdaptedResult) {
+	if e.defaultTracer != nil && e.cfg.AdapterConfig.NeedTraceAsResourceSpan {
+		_, span := e.defaultTracer.Start(
+			context.Background(),
+			constvalues.SpanInfo,
+			apitrace.WithAttributes(result.Attrs...),
+		)
+		span.End()
+	} else if e.defaultTracer == nil && e.cfg.AdapterConfig.NeedTraceAsResourceSpan {
```
dxsup, line 66: If and else if can be simplified by a guard clause.
```go
if !e.cfg.AdapterConfig.NeedTraceAsResourceSpan {
    return
}
if e.defaultTracer == nil {
    // Log something
    return
}
// Normal behavior
...
```

NeJan2020: done

collector/consumer/exporter/otelexporter/instrument.go
NeJan2020: By expanding the gauge into a gauge group, `recordLastValue()` is using a single DefaultAggregator now

collector/consumer/exporter/otelexporter/consume.go
NeJan2020: The task of modifying the final output metric's name is completed by the adapter now

collector/consumer/exporter/otelexporter/defaultadapter/adapter.go
```diff
@@ -0,0 +1,57 @@
...
+	Metric ResultType = "metric"
+	Trace             = "trace"
+)
+
+type RenameRule int
+
+const (
+	ServerMetrics RenameRule = iota
+	TopologyMetrics
+	KeepOrigin
```
dxsup, line 47: Please remove these unused codes.

collector/pkg/aggregator/defaultaggregator/default_aggregator.go
```diff
@@ -55,3 +55,22 @@ func (s *DefaultAggregator) Dump() []*model.GaugeGroup {
...
+		result := vr.dump()
+		vr.reset()
+		return result
+	}
+	return nil
+}
+
+func (s *DefaultAggregator) CheckExist(gaugeGroupName string) bool {
+	s.mut.Lock()
+	defer s.mut.Unlock()
```
dxsup, line 73: The `Lock` can be replaced by `RLock`.

collector/consumer/exporter/otelexporter/defaultadapter/simple_adapter.go
```diff
@@ -0,0 +1,52 @@
...
+
+type SimpleAdapter struct {
+	acceptGaugeGroupNames map[string]struct{}
+	constLabels           []attribute.KeyValue
+}
+
+func (d *SimpleAdapter) Adapt(gaugeGroup *model.GaugeGroup) ([]*AdaptedResult, error) {
+	if _, ok := d.acceptGaugeGroupNames[gaugeGroup.Name]; ok {
+		return []*AdaptedResult{
+			{
```
dxsup, line 16: Better if use guard clause.

collector/consumer/exporter/otelexporter/instrument.go
```diff
@@ -2,8 +2,14 @@ package otelexporter
...
 import (
 	"context"
+	"errors"
+	"fmt"
+	"github.com/Kindling-project/kindling/collector/consumer/exporter/otelexporter/defaultadapter"
 	"github.com/Kindling-project/kindling/collector/model"
 	"github.com/Kindling-project/kindling/collector/model/constlabels"
+	"github.com/Kindling-project/kindling/collector/model/constnames"
+	"github.com/Kindling-project/kindling/collector/pkg/aggregator"
+	defaultaggregator "github.com/Kindling-project/kindling/collector/pkg/aggregator/defaultaggregator"
```
dxsup, line 12: Redundant alias which could be removed.

dxsup (dismissed):
LGTM

dxsup (approved):

Issue #194, opened 2022-04-22
Concurrent problems in default aggregator module


dxsup:
**Describe the bug**
The result value calculated by the default aggregator is wrong when aggregating and dumping are executed concurrently.  

**How to reproduce?**
I can always reproduce this problem by running the following test case.
```go

func TestConcurrentAggregator(t *testing.T) {
	aggKindMap := &AggregatedConfig{KindMap: map[string][]KindConfig{
		"duration": {
			{Kind: SumKind, OutputName: "duration_sum"},
			{Kind: CountKind, OutputName: "request_count"},
		},
	}}

	aggregator := NewDefaultAggregator(aggKindMap)
	labels := model.NewAttributeMap()
	labels.AddStringValue("key", "value")

	labelSelectors := internal.NewLabelSelectors(
		internal.LabelSelector{Name: "key", VType: internal.StringType},
	)

	wg := sync.WaitGroup{}
	wg.Add(1)
	stopCh := make(chan bool)

	var runLoop = 100000
	var duration int64 = 100
	go func() {
		for i := 0; i < runLoop; i++ {
			gaugeValues := []*model.Gauge{
				{"duration", duration},
			}
			gaugeGroup := model.NewGaugeGroup("testGauge", labels, 0, gaugeValues...)
			aggregator.Aggregate(gaugeGroup, labelSelectors)
			time.Sleep(time.Microsecond)
		}
		stopCh <- true
		wg.Done()
	}()

	wg.Add(1)
	var durationSum int64
	var requestCount int64
	go func() {
		ticker := time.NewTicker(500 * time.Microsecond)
		for {
			select {
			case <-ticker.C:
				ret := aggregator.Dump()
				for _, g := range ret {
					for _, v := range g.Values {
						if v.Name == "duration_sum" {
							durationSum += v.Value
						} else if v.Name == "request_count" {
							requestCount += v.Value
						}
					}
				}
			case <-stopCh:
				ret := aggregator.Dump()
				for _, g := range ret {
					for _, v := range g.Values {
						if v.Name == "duration_sum" {
							durationSum += v.Value
						} else if v.Name == "request_count" {
							requestCount += v.Value
						}
					}
				}
				wg.Done()
				return
			}
		}
	}()
	wg.Wait()
	expectedDurationSum := int64(runLoop) * duration
	expectedRequestCount := int64(runLoop)
	if durationSum != expectedDurationSum || requestCount != expectedRequestCount {
[...]

PR #195, opened 2022-04-22
Add concurrency constraints to the defaultaggregator

dxsup:
## Description

1. RLock when calling the `Aggregate()` method and lock when calling the `Dump()` method.
2. Use `sync.Map` to replace the native map with a mutex to improve readability.

## Related Issue

#194 

## How Has This Been Tested?

Run the same test case with #194, and never see the error again.
=== RUN   TestConcurrentAggregator
--- PASS: TestConcurrentAggregator (1.66s)
PASS

### DIFF
--- a/collector/consumer/processor/aggregateprocessor/internal/defaultaggregator/value_recorder.go
+++ b/collector/consumer/processor/aggregateprocessor/internal/defaultaggregator/value_recorder.go
@@ -50,8 +50,8 @@ func (r *valueRecorder) Record(key *internal.LabelKeys, gaugeValues []*model.Gau
 // dump a set of metric from counter cache.
 // The return value holds the reference to the metric, not the copied one.
 func (r *valueRecorder) dump() []*model.GaugeGroup {
-	r.mutex.RLock()
-	defer r.mutex.RUnlock()
+	r.mutex.Lock()
+	defer r.mutex.Unlock()
 	ret := make([]*model.GaugeGroup, len(r.labelValues))
 	index := 0
 	for k, v := range r.labelValues {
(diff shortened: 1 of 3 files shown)


collector/consumer/processor/aggregateprocessor/internal/defaultaggregator/value_recorder.go
```diff
@@ -29,41 +26,32 @@ func (r *valueRecorder) Record(key *internal.LabelKeys, gaugeValues []*model.Gau
...
-		r.mutex.Lock()
 		// double check to avoid double writing
-		aggValues, ok = r.labelValues[*key]
-		if !ok {
-			aggValues = newAggValuesMap(gaugeValues, r.aggKindMap)
-			r.labelValues[*key] = aggValues
-		}
-		r.mutex.Unlock()
+		aggValues, _ = r.labelValues.LoadOrStore(*key, newAggValuesMap(gaugeValues, r.aggKindMap))
 	}
```
NeJan2020, line 33: No double check anymore

NeJan2020 (approved):
LGTM, Although the `r.labelValues = sync.Map{}` in `valueRecorder.reset` has a bad smell

collector/consumer/processor/aggregateprocessor/internal/defaultaggregator/value_recorder.go
dxsup: `LoadOrStore` is actually the "double-check", otherwise `Store` will be used.

dxsup:
> LGTM, Although the `r.labelValues = sync.Map{}` in `valueRecorder.reset` has a bad smell

Yes, it would cause a fatal error if using the `sync.Map` while resetting it. Any good ideas?

PR #196, opened 2022-04-24
Add the description of NOT_FOUND to the metric doc

### LINKED ISSUE #137: Confusion about NOT_FOUND_EXTERNAL and NOT_FOUND_INTERNAL IP types
In the topology view, I found IP node tags with two types: NOT_FOUND_EXTERNAL and NOT_FOUND_INTERNAL.  What do those types stand for and why Kindling provide two different IP types.

image

dxsup:
## Description

Write down what `NOT_FOUND_INTERNAL` and `NOT_FOUND_EXTENRAL`  mean on the Prometheus metrics docs.

## Related Issue

This will close #137

### DIFF
--- a/docs/prometheus_metrics.md
+++ b/docs/prometheus_metrics.md
@@ -14,7 +14,7 @@ Service metrics are generated from the server-side events, which are used to sho
 | **Label Name** | **Example** | **Notes** |
 | --- | --- | --- |
 | node | worker-1 | Node name represented in Kubernetes cluster |
-| namespace | default | Namespace of the Pod |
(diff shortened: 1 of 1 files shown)


J-lena (changes requested):
I think the label "NOT_FOUND" should be described more accurately. Assuming that the user has installed Kindling in two Kubernetes clusters, If one cluster directly accesses the POD IP (in MacVLAN Network mode) of the other cluster, the POD IP will also be labeled NOT_FOUND. So the process scope of NOT_FOUND is within a single K8S cluster.

J-lena (approved):

Issue #197, opened 2022-04-25
No more kindling events can be sent after the convert list is full


sanyangji:
**Describe the bug**
No more kindling events can be sent after the convert list exceeds the max size. The process of convert stopped when list is full because it has no try to swap convert list and send list.

**How to reproduce?**
1. set list_max_size.
2. give some pressure to exceed the max size.
3. get the convert part stopped which causes no more kindling events can be sent.

**What did you expect to see?**
kindling events can be sent correctly.

dxsup:
Closed by #200

PR #200, opened 2022-04-27
Fix: convert stopped when list is full

sanyangji:
## Related Issue
#197 

## Motivation and Context
fix: convert stopped when list is full

## How Has This Been Tested?
1. set list_max_size to X
2. make the convert list size to X continuously
3. check the convert part is running normally and the max size of convert list is X.

### DIFF
--- a/probe/src/probe/publisher/publisher.cpp
+++ b/probe/src/probe/publisher/publisher.cpp
@@ -40,9 +40,6 @@ void publisher::consume_sysdig_event(sinsp_evt *evt, int pid, converter *sysdigC
     }
     // convert sysdig event to kindling event
     if (m_selector->select(evt->get_type(), ((sysdig_converter *) sysdigConverter)->get_kindling_category(evt))) {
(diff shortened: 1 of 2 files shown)


dxsup (dismissed):

probe/src/probe/publisher/publisher.cpp
```diff
@@ -52,12 +49,21 @@ void publisher::consume_sysdig_event(sinsp_evt *evt, int pid, converter *sysdigC
         } else {
             kindlingEventList = it->second;
         }
+
+        if (sysdigConverter->judge_max_size()) {
+            // check if the send list has sent
+            if (!m_ready[kindlingEventList]) {
```
jundizhou, line 55:
```
            if (m_ready[kindlingEventList]) {
                 // drop event
                return;
            } 
           swap_list(sysdigConverter, kindlingEventList);
```
It's better to write like this

jundizhou (approved):

PR #204, opened 2022-04-29
Remove format-processor from application and config file

NeJan2020:
Since this component has been replaced by the default adapter in OTEL-exporter, the remaining code needs to be deleted.

This PR will also remove the config of the format-processor from the config file.


dxsup (approved):
LGTM

PR #207, opened 2022-05-05
fix(exporter): use int to store dst_port and dnat_port

NeJan2020:
1. add a test case in label_converter_test to check this bug
2. add a test case in instrument_test to test the aggregator


dxsup (approved):
LGTM

PR #210, opened 2022-05-06
Print probe version

sugary199:
Print probe version in both terminal and log. 
## Description

Generated a header file `version.h`with `probe-build.sh`
Reference it and output in the main function
## Related Issue

https://github.com/Kindling-project/kindling/issues/112
## How Has This Been Tested?

you can see that in probe container terminal
```
KINDLING_PROBE_version: 48381d3 commit: 2022-04-26 17:37:44 build: 2022-04-27 23:42:56
```

### DIFF
--- a/probe/src/probe/BUILD.bazel
+++ b/probe/src/probe/BUILD.bazel
@@ -81,9 +81,9 @@ container_push(
     format = "Docker",
     image = ":kindling_probe_image",
     # ATTENTION!! Modify this if you want to change docker registry.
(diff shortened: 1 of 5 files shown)


probe/src/probe/BUILD.bazel
```diff
@@ -80,7 +80,7 @@ container_push(
     name = "push_image",
     format = "Docker",
```
jundizhou, line 81: Please do not modify here

PR #212, opened 2022-05-07
fix/use `UNSUPPORT` to represent undefined procotol in labelConverter

NeJan2020:
1.  fix the exception when sending `Kafka` or `gRPC` trace data  as Span through otlp-grpc.
2. fix the exception when sending trace as Span with prometheus.


collector/consumer/exporter/otelexporter/defaultadapter/label_converter.go
```diff
@@ -313,14 +313,25 @@ func (m *LabelConverter) transform(group *model.GaugeGroup) (*model.AttributeMap
...
 	for i := 0; i < len(m.updateKeys); i++ {
 		tmpExtraKey = m.updateKeys[i](tmpExtraKey, labels)
 	}
-	attrs := m.labelsMap[*tmpExtraKey]
+	attrs, ok := m.labelsMap[*tmpExtraKey]
+	if !ok {
+		tmpExtraKey.protocol = UNSUPPORTED
+		attrs = m.labelsMap[*tmpExtraKey]
+	}
+	return attrs
```
dxsup, line 328: Better if we add a default result value when the `attrs` cannot be found even if using `UNSUPPORTED`. Since currently `UNSUPPORTED` always resides on the map, the modification could be made later.

dxsup (approved):
LGTM

PR #220, opened 2022-05-13
modify grafana dashboard

thousandxu:
## Description
1.Modify the field display of the table
2.Modify interval parameter for Prometheus Query in line、bar and pie chart

## Motivation and Context
Solve the problem of inaccurate chart data。


dxsup (approved):
LGTM

PR #223, opened 2022-05-13
counting the page-faults (based on switch)

yaofighting:

## Description

Add the function of counting page fault major and page fault minor (obtained based on switch event)
Interpretation of the meaning of metrcs:
1. kindling_pagefault_major_total: it corresponds to the number of major page faults captured by the kernel. The specific statistical value is provided by the kernel and associated with specific threads. (eg. when the page to be accessed by the process is not in the real physical memory, it may need to be loaded from the disk (for example, Linux turns on the swap mechanism, and the kernel replaces the page to the disk for temporary storage through the LRU page replacement algorithm), a major page fault will be generated at this time)
2. kindling_pagefault_minor_total: it corresponds to the number of minor page faults captured by the kernel. The specific statistical value is provided by the kernel and associated with specific threads. (eg. when the page to be accessed by the process is in the physical memory, but the MMU has not established the mapping relationship between the virtual address of the process and the corresponding physical address, the minor page fault will be triggered)

## Motivation and Context

We need kindling support to get the data information of user thread page fault

## How Has This Been Tested?

The subscription of switch events is added, and the part about page faults analyzer and peocessor pipeline processing in connector is added. Finally, the data is exported through otelexporter


dxsup:
Please add you changes to the configuration file at `collector/docker/kindling-collector-config.yml`, too.

collector/docker/kindling-collector-config.yml
```diff
@@ -110,6 +112,8 @@ exporters:
       # send trace data in the format of ResourceSpan
       need_trace_as_span: false
     metric_aggregation_map:
+      kindling_pagefault_major_total: gauge
+      kindling_pagefault_minor_total: gauge
```
NeJan2020, line 116: In Prometheus, `gauge` is used to record non monotonic data, corresponding to `gaugeObserver` in OTEL. (check this in ins.recordLastValue()). Obviously, the page fault metric is monotonically increasing and cumulative pre-aggregated metric, which corresponds to `counterObserver` in OTEL. Can you add some support for this kind of data?

dxsup:
Could you add your metric description to the [prometheus_metrics.md](https://github.com/CloudDectective-Harmonycloud/kindling/blob/main/docs/prometheus_metrics.md)?

collector/analyzer/pgftmetricanalyzer/pgft_analyzer.go
```diff
@@ -0,0 +1,157 @@
...
+}
+
+func (a *PgftMetricAnalyzer) getSwitchLabels(event *model.KindlingEvent) (*model.AttributeMap, error) {
+
+	// next := event.GetUserAttribute("next")
+	//pgftMaj := event.GetUserAttribute("pgft_maj")
+	//pgftMin := event.GetUserAttribute("pgft_min")
+	// vmSize := event.GetUserAttribute("vm_size")
+	// vmRss := event.GetUserAttribute("vm_rss")
+	// vmSwap := event.GetUserAttribute("vm_swap")
```
dxsup, line 118: Please remove these unused/commented codes to make things clear.

collector/consumer/processor/aggregateprocessor/processor.go
```diff
@@ -146,7 +152,27 @@ func newNetRequestLabelSelectors() *aggregator.LabelSelectors {
...
+		aggregator.LabelSelector{Name: constlabels.Ip, VType: aggregator.StringType},
+		aggregator.LabelSelector{Name: constlabels.Service, VType: aggregator.StringType},
+		aggregator.LabelSelector{Name: constlabels.Node, VType: aggregator.StringType},
+		aggregator.LabelSelector{Name: constlabels.Namespace, VType: aggregator.StringType},
+		//aggregator.LabelSelector{Name: constlabels.NextPid, VType: aggregator.IntType},
+		//aggregator.LabelSelector{Name: constlabels.PgftMaj, VType: aggregator.IntType},
+		//aggregator.LabelSelector{Name: constlabels.PgftMin, VType: aggregator.IntType},
+		//aggregator.LabelSelector{Name: constlabels.VmSize, VType: aggregator.IntType},
+		//aggregator.LabelSelector{Name: constlabels.VmRss, VType: aggregator.IntType},
+		//aggregator.LabelSelector{Name: constlabels.VmSwap, VType: aggregator.IntType},
```
dxsup, line 169: Remove these commented codes.

collector/docker/kindling-collector-config.yml
dxsup: It's tough work for him to add this support. We better leave this behind and might do it later. Prometheus supports "increase()" or "rate()" on this kind of metric that is `gauge` but monotonically increasing, so I think `gauge` is ok for this scenario.

collector/analyzer/pgftmetricanalyzer/pgft_analyzer.go
```diff
@@ -0,0 +1,137 @@
...
+	PgftMetric analyzer.Type = "pgftmetricanalyzer"
+)
+
+var consumableEvents = map[string]bool{
+	constnames.SwitchEvent: true,
+}
+
+type PgftMetricAnalyzer struct {
+	consumers   []consumer.Consumer
+	conntracker conntrackerpackge.Conntracker
```
dxsup, line 26: You won't need `conntracker` since your metric is none of business with the network.

collector/analyzer/pgftmetricanalyzer/pgft_analyzer.go
```diff
@@ -0,0 +1,137 @@
...
+
+	return model.NewGaugeGroup(constnames.PgftGaugeGroupName, labels, event.Timestamp, gaugeSlice...), nil
+}
+
+func (a *PgftMetricAnalyzer) getSwitchLabels(event *model.KindlingEvent) (*model.AttributeMap, error) {
+
+	ctx := event.GetCtx()
+	threadinfo := ctx.GetThreadInfo()
+	tid := (int64)(threadinfo.GetTid())
+	pid := (int64)(threadinfo.GetPid())
```
dxsup, line 116: Is it possible that `ctx` or `threadinfo` is nil? If yes, this will panic the program.

collector/docker/kindling-collector-config.yml
NeJan2020: I will open a new issue for this later.

probe/src/probe/publisher/publisher.cpp
```diff
@@ -27,6 +28,14 @@ publisher::~publisher() {
     delete m_client_event_map;
 }
 
+bool filterSwitch(char *val, int threshold){
+    int len = strlen(val), num = 0;
+    for(int i = 0;i < len;i++){
+        num *= 10;
+        num += val[i] - '0';
+    }
```
dxsup, line 36: I don't think it is a good idea to write your own function to convert `char*` to `int` since there is a simple function `atoi` in the standard library which can simplify your codes and ensure the logic is correct. I'm really concerned about these non-test codes, especially when it is wrote in C++.

collector/analyzer/pgftmetricanalyzer/pgft_analyzer.go
```diff
@@ -110,14 +110,23 @@ func (a *PgftMetricAnalyzer) generateSwitchPgft(event *model.KindlingEvent) (*mo
 
 func (a *PgftMetricAnalyzer) getSwitchLabels(event *model.KindlingEvent) (*model.AttributeMap, error) {
 
+	labels := model.NewAttributeMap()
 	ctx := event.GetCtx()
+	if ctx == nil {
+		return labels, nil
+	}
+
```
dxsup, line 111: It's typically that we return an error when the return value doesn't contain useful information. As the metrics are meaningless if the labels are empty, we better return an error and print a debug log outside this method to trace the data, which has been done in method `ConsumeEvent` (line 52).

NeJan2020 (approved):
LGTM

NeJan2020 (approved):

dxsup:
This PR can't be merged due to the barrier to performance. Using the `switch` event to implement the count of `page fault` could significantly impact the performance. I know you are working on another implementation, and once it is done you can open another PR. Therefore, I think this PR should be closed now.

Issue #224, opened 2022-05-16
the collector start failed caused by "stat ~/.kube/config: no such file or directory"


yaofighting:
**Describe the bug**
when I attempt to run the collector at centOS 7.5, I find an error about "stat ~/.kube/config: no such file or directory"
**How to reproduce?**
run the collector process at centOS 7.5

**What did you expect to see?**
the collector should run normally

**What did you see instead?**
the collector didn't run normally with the error infomation "stat ~/.kube/config: no such file or directory"

**Screenshots**
![image](https://user-images.githubusercontent.com/46129108/168530686-93254a29-18be-4aa0-8dc2-6e58e87b2bbd.png)

**What config did you use?**
Config: kindling-collector-config.yml

**Logs**
Please attach the logs by running the following command:
```
kubectl logs -f kindling-agent-xxx -n kindling -c kindling-probe
kubectl logs -f kindling-agent-xxx -n kindling -c kindling-collector
```
![image](https://user-images.githubusercontent.com/46129108/168532383-ccf750dc-a9ac-458b-aa84-5fbacd5c9783.png)

**Environment (please complete the following information)**
- Kindling agent version
- Kindlinng-falcon-lib version
- Node OS version
- Node Kernel version
- Kubernetes version
- Prometheus version
- Grafana version

NeJan2020:
Closed by #225

PR #225, opened 2022-05-16
fix the kindling-collector-config.yml caused by "stat ~/.kube/config: no such file or directory"

yaofighting:

fix the collector config and make it running normally on the Linux environment
## Description

in the file /root/kindling/collector/docker/kindling-collector-config.yml, we can see a bug about that we may not boot the collector normally in the Linux environment due to the "kube_config_dir" is "~" instead of "root", so in this PR, I fix it.
## Related Issue

https://github.com/Kindling-project/kindling/issues/224
## Motivation and Context

it fix the compatibility of collector
## How Has This Been Tested?

just compile and build the collector container with the doc's step.

### DIFF
--- a/collector/docker/kindling-collector-config.yml
+++ b/collector/docker/kindling-collector-config.yml
@@ -73,7 +73,7 @@ analyzers:
 processors:
   k8smetadataprocessor:
     kube_auth_type: kubeConfig
-    kube_config_dir: ~/.kube/config
+    kube_config_dir: root/.kube/config
     grace_delete_period: 60
   aggregateprocessor:
     # Aggregation duration window size. The unit is second.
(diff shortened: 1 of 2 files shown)


collector/docker/kindling-collector-config.yml
```diff
@@ -73,7 +73,7 @@ analyzers:
 processors:
   k8smetadataprocessor:
     kube_auth_type: kubeConfig
-    kube_config_dir: ~/.kube/config
+    kube_config_dir: root/.kube/config
```
dxsup, line 76: Should be started with "/".

dxsup (changes requested):
There is another configuration file under the directory `kindling/deploy`. Although it is using `serviceaccount` as its `kube_auth_type` and `kube_config_dir` isn't effective, I think we better make this field same to avoid misunderstanding.

collector/docker/kindling-collector-config.yml
yaofighting: OK, it has been modified as required

dxsup (approved):
LGTM

PR #232, opened 2022-06-01
Decouple the logic of dispatching events from receivers

dxsup:
## Description

I have removed the logic of dispatching events from receivers so that now receivers don't care about the analyzers after the initialization. All changes about dispatching events are limited in the analyzers. 
AnalyzerManager now takes a more important responsibility for creating the dispatching logic.

## Motivation and Context

In the current implementation, we can't send an event to different analyzers, which is necessary for implementing another metric(TCP connect metrics which I'm working on). This PR allows the receivers to do that.


collector/analyzer/manager_test.go
```diff
@@ -0,0 +1,86 @@
...
+	assert.True(t, ok)
+	_, ok = analyzers[1].(*testConsumeAllAnalyzer)
+	assert.True(t, ok)
+
+	analyzers = manager.GetConsumableAnalyzers("evt2")
+	assert.Equal(t, 2, len(analyzers))
+	_, ok = analyzers[0].(*testAnalyzer)
+	assert.True(t, ok)
+	_, ok = analyzers[1].(*testConsumeAllAnalyzer)
+	assert.True(t, ok)
```
NeJan2020, line 29: Here needs a test case using undefined evt_name， such as "evt3"

collector/analyzer/tcpmetricanalyzer/tcp_analyzer.go
```diff
@@ -49,6 +49,14 @@ func (a *TcpMetricAnalyzer) Start() error {
 	return nil
 }
 
+func (a *TcpMetricAnalyzer) ConsumableEvents() []string {
+	ret := make([]string, len(consumableEvents))
+	for key := range consumableEvents {
+		ret = append(ret, key)
+	}
```
NeJan2020, line 56: Why not change consumableEvents into a string list and just remove the double-check in the `ConsumeEvent`?

dxsup: Good idea. Done.

NeJan2020 (approved):

PR #233, opened 2022-06-01
Search for k8s metadata using src_ip when no containerid found

dxsup:
## Description

Search for k8s metadata using `src_ip` when no `container_id` is found.

## Motivation and Context

We have found that the event from the client-side doesn't contain the container id in a very small possibility, in which case we take the src_ip as the key to find its k8s metadata.

### DIFF
--- a/collector/consumer/processor/k8sprocessor/kubernetes_processor.go
+++ b/collector/consumer/processor/k8sprocessor/kubernetes_processor.go
@@ -1,6 +1,8 @@
 package k8sprocessor
 
 import (
+	"strconv"
+
 	"github.com/Kindling-project/kindling/collector/component"
 	"github.com/Kindling-project/kindling/collector/consumer"
 	"github.com/Kindling-project/kindling/collector/consumer/processor"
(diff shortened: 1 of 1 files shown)


collector/consumer/processor/k8sprocessor/kubernetes_processor.go
```diff
@@ -85,15 +86,36 @@ func (p *K8sMetadataProcessor) processTcpMetric(dataGroup *model.DataGroup) {
...
+		}
 	} else {
-		labelMap.UpdateAddStringValue(constlabels.SrcNodeIp, p.localNodeIp)
-		labelMap.UpdateAddStringValue(constlabels.SrcNode, p.localNodeName)
-		labelMap.UpdateAddStringValue(constlabels.SrcNamespace, constlabels.InternalClusterNamespace)
+		srcIp := labelMap.GetStringValue(constlabels.SrcIp)
+		if srcIp == loopbackIp {
+			labelMap.UpdateAddStringValue(constlabels.SrcNodeIp, p.localNodeIp)
+			labelMap.UpdateAddStringValue(constlabels.SrcNode, p.localNodeName)
+		}
```
NeJan2020, line 104: LGTM, Although The k8s information searched may be incorrect when processing data from the loopback device.

NeJan2020 (approved):

PR #238, opened 2022-06-06
Set container name empty when multiple containers don't specify ports

dxsup:
## Description

Set container name empty when multiple containers don't specify ports.

## Motivation and Context

When there are multiple containers specified in one pod but they don't specify ports listened, we added them to the map using the combination of pod IP and port 0 as the key. This would cause us to find an incorrect container name using pod IP and the listening port. Since there is no method to find the exact container we want, I think we should set the container name to empty in case it makes the final result confusing.

Note that if there is only one container in the pod and it doesn't specify its port, we remain its name since no other containers confuse us.


NeJan2020 (approved):
This fix also influence ISSUE #239

PR #240, opened 2022-06-07
Block in the application instead of the udsreceiver after running 

dxsup:
## Description

1. Block the program in the main function instead of the udsreceiver after running. This is necessary if we want to replace the current receiver, otherwise, the application will exit immediately.
2. Fix the typo `shutdwonState` to `shutdownState` in the udsreceiver.
3. Set receiving high water mark of the pull socket instead of sending.

### DIFF
--- a/collector/application/application.go
+++ b/collector/application/application.go
@@ -3,7 +3,6 @@ package application
 import (
 	"flag"
 	"fmt"
-	"sync"
 
 	"github.com/Kindling-project/kindling/collector/analyzer"
 	"github.com/Kindling-project/kindling/collector/analyzer/loganalyzer"
@@ -28,7 +27,6 @@ type Application struct {
 	telemetry         *component.TelemetryManager
 	receiver          receiver.Receiver
 	analyzerManager   *analyzer.Manager
-	shutdownWG        sync.WaitGroup
 }
 
 func New() (*Application, error) {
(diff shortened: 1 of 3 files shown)


collector/application/application.go
```diff
@@ -58,14 +60,18 @@ func (a *Application) Run() error {
...
 	}
+	a.shutdownWG.Wait()
 	return nil
 }
 
 func (a *Application) Shutdown() error {
-	return multierr.Combine(a.receiver.Shutdown(), a.analyzerManager.ShutdownAll(a.telemetry.Telemetry.Logger))
+	err := a.receiver.Shutdown()
+	a.shutdownWG.Done()
+	return multierr.Combine(err, a.analyzerManager.ShutdownAll(a.telemetry.Telemetry.Logger))
```
NeJan2020, line 74: `a.analyzerManager.ShutdownAll` may not execute normally since the main thread has already exited after `a.shutdownWG.Done()`.

NeJan2020, line 74: `a.analyzerManager.ShutdownAll` may not execute normally since the main thread has already exited after `a.shutdownWG.Done()`.

collector/receiver/udsreceiver/udsreceiver.go
```diff
@@ -122,12 +122,12 @@ func (r *UdsReceiver) startZeroMqPull() error {
 	go func() {
 		defer r.shutdownWG.Done()
```
NeJan2020, line 123: Maybe remove this `sync.WaitGroup` since we are not using them?

collector/application/application.go
dxsup: Oh, you are right. I didn't call the `Shutdown` method so I forgot to check it. I have moved the blocking part to the main function.

NeJan2020 (approved):

PR #243, opened 2022-06-09
feat:add some self metric for agent cpu and memory usage

NeJan2020:
## Description
This PR adds some extra performance metrics, which could be used in self-monitoring or e2e tests.

Extra Metrics
- kindling_telemetry_agent_cpu_time_seconds 
  - CPU time cost by agent(collector) in system and user , the unit is second, differentiated by the label named 'type'
- kindling_telemetry_agent_memory_used_bytes
  - Resident set size (RSS) or Virtual memory size (VMS) occupied by agent(collectir)

### DIFF
--- a/collector/go.mod
+++ b/collector/go.mod
@@ -14,9 +14,12 @@ require (
 	github.com/pebbe/zmq4 v1.2.7
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.11.0 // indirect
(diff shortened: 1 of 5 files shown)


dxsup:
Could you explain why these metrics are necessary since cAdvisor has observed them?

dxsup (approved):
LGTM

dxsup (approved):
LGTM

PR #250, opened 2022-06-14
Always add duration metric to tcp_connect_data_group

dxsup:
## Description

Always add duration metric to `tcp_connect_data_group`.

## Motivation and Context

The structure `DataGroup` is used to group a set of metrics that contain the same labels. Any group should have a clear definition of its values and labels. The `tcp_connect_data_group` should always contain two values but has missed the `duration` value when the connection failed to be established. This makes it inconvenient for users willing to process the data further since they cannot make sure what values it contains.

### DIFF
--- a/collector/analyzer/tcpconnectanalyzer/analyzer.go
+++ b/collector/analyzer/tcpconnectanalyzer/analyzer.go
@@ -182,6 +182,8 @@ func (a *TcpConnectAnalyzer) generateDataGroup(connectStats *internal.Connection
 	// Only record the connection's duration when it is successfully established
 	if connectStats.StateMachine.GetCurrentState() == internal.Success {
 		metrics = append(metrics, model.NewIntMetric(constnames.TcpConnectDurationMetric, connectStats.GetConnectDuration()))
+	} else {
+		metrics = append(metrics, model.NewIntMetric(constnames.TcpConnectDurationMetric, 0))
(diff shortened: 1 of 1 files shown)


collector/analyzer/tcpconnectanalyzer/analyzer.go
```diff
@@ -182,6 +182,8 @@ func (a *TcpConnectAnalyzer) generateDataGroup(connectStats *internal.Connection
 	// Only record the connection's duration when it is successfully established
 	if connectStats.StateMachine.GetCurrentState() == internal.Success {
 		metrics = append(metrics, model.NewIntMetric(constnames.TcpConnectDurationMetric, connectStats.GetConnectDuration()))
+	} else {
+		metrics = append(metrics, model.NewIntMetric(constnames.TcpConnectDurationMetric, 0))
```
NeJan2020, line 186: Such changes can seriously affect the average connection time. How about setting this value to an impossible value and adding something special logic when aggregating?

dxsup: This is a serious problem I didn't notice before. Adding special logic when aggregating is not a good idea as aggregating processor shouldn't know too much about the data itself. There isn't an appropriate method that arose in my mind to fix this issue now. I will close this PR temporarily.

PR #251, opened 2022-06-16
k8sprocessor: use src IP for further searching if the dst IP is a loopback address

dxsup:
## Description
Use src IP for further searching if the dst IP is a loopback address.

## Motivation and Context

I found there are some cases where the containers use the loopback address to send requests to the containers in the same pod, such as `node-exporter`. In these cases, the destination IP is `127.0.0.1` with which we can't find the pod information, but the source IP is correct which can be used to search for the destination pod.


NeJan2020 (approved):
This change will not affect the original data(the 'dstIp' field) but simply uses reasoning to determine the true `dstPod` information. We need to modify some of the existing logic and maintain this principle in the future, that is any reasoning should not change the origin data.

Issue #142, opened 2022-03-24
Wrong usage of zap.logger.debug


NeJan2020:
**Describe the bug**
In the performance test, we found a huge cost by `String()`, which was caused by the lines below.

```go
p.telemetry.Logger.Debug("dstNodeIp or srcNodeIp is empty which is not expected, skip: ", zap.String("gaugeGroup", gaugeGroup.String()))
```
Here is the flame graph.

<img width="1411" alt="image" src="https://user-images.githubusercontent.com/64676199/159875495-03439ba9-7254-4bf1-bc19-f27f3cd6ba18.png">

To avoid this cost, When we use `zap.logger.debug` to add some logs in `debug` level, we should use the `zap.logger.Check()` function, like blow

```go
if ce := p.telemetry.Logger.Check(zapcore.DebugLevel, "dstNodeIp or srcNodeIp is empty which is not expected, skip: "); ce != nil {
	ce.Write(zap.String("gaugeGroup", gaugeGroup.String()))
}
```

I will add a PR to fix this kind of problem in our project now.

PR #253, opened 2022-06-16
Remove the histogram metrics by default

### LINKED ISSUE #252: Remove the histogram metrics by default
Several users have reported that our two metrics `kindling_entity_request_average_duration_nanoseconds` and `kindling_topology_request_average_duration_nanoseconds` have the highest cardinality in their Prometheus, which causes a high load on both their Prometheus and our agents. 

This is one of the Prometheus servers.
high-cardinality

And another one.
Tcp 10 Series count bor metric names

I have checked the grafana-plugins and found these two metrics are not used now. So I think we better remove them by default.

dxsup:
## Description

1. Remove the related lines in the configuration file.
2. Modify the metrics description document.
3. Lower some log levels of the otelexporter to avoid printing too many logs.

## Related Issue

This PR will close #252

### DIFF
--- a/collector/consumer/exporter/otelexporter/consume.go
+++ b/collector/consumer/exporter/otelexporter/consume.go
@@ -11,6 +11,7 @@ import (
 	"go.opentelemetry.io/otel/metric"
 	apitrace "go.opentelemetry.io/otel/trace"
 	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
 )
 
 func (e *OtelExporter) Consume(dataGroup *model.DataGroup) error {
@@ -90,7 +91,9 @@ func (e *OtelExporter) exportMetric(result *adapter.AdaptedResult) {
 		} else if metric.DataType() == model.HistogramMetricType {
 			e.telemetry.Logger.Warn("Failed to exporter Metric: can not use otlp-exporter to export histogram Data", zap.String("MetricName", metric.Name))
 		} else {
-			e.telemetry.Logger.Debug("Undefined metricKind for this Metric", zap.String("MetricName", metric.Name), zap.String("MetricType", reflect.TypeOf(metric).String()))
+			if ce := e.telemetry.Logger.Check(zapcore.DebugLevel, "Undefined metricKind for this Metric"); ce != nil {
+				ce.Write(zap.String("MetricName", metric.Name), zap.String("MetricType", reflect.TypeOf(metric).String()))
+			}
 		}
 	}
 	if len(measurements) > 0 {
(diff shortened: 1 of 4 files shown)


deploy/agent/kindling-collector-config.yml
```diff
@@ -122,12 +122,10 @@ exporters:
     metric_aggregation_map:
       kindling_entity_request_total: counter
       kindling_entity_request_duration_nanoseconds_total: counter
-      kindling_entity_request_average_duration_nanoseconds: histogram
       kindling_entity_request_send_bytes_total: counter
       kindling_entity_request_receive_bytes_total: counter
       kindling_topology_request_total: counter
       kindling_topology_request_duration_nanoseconds_total: counter
-      kindling_topology_request_average_duration_nanoseconds: histogram
```
NeJan2020, line 130: Please also remove the average time from aggregate_kind_map, otherwise, the collector may print logs of 'Undefined metricKind' in the log

dxsup: I have lowered the log level to `debug` to avoid this problem. It will be too inconvenient for users to add them back if we remove the average time from aggregate_kind_map, so I prefer to remaining it.

collector/consumer/exporter/otelexporter/consume.go
```diff
@@ -88,9 +88,9 @@ func (e *OtelExporter) exportMetric(result *adapter.AdaptedResult) {
 		} else if ok && metric.DataType() == model.IntMetricType {
 			measurements = append(measurements, e.instrumentFactory.getInstrument(metric.Name, metricKind).Measurement(metric.GetInt().Value))
 		} else if metric.DataType() == model.HistogramMetricType {
-			e.telemetry.Logger.Error("Failed to exporter Metric: can not use otlp-exporter to export histogram Data", zap.String("MetricName", metric.Name))
+			e.telemetry.Logger.Warn("Failed to exporter Metric: can not use otlp-exporter to export histogram Data", zap.String("MetricName", metric.Name))
 		} else {
-			e.telemetry.Logger.Warn("Undefined metricKind for this Metric", zap.String("MetricName", metric.Name), zap.String("MetricType", reflect.TypeOf(metric).String()))
+			e.telemetry.Logger.Debug("Undefined metricKind for this Metric", zap.String("MetricName", metric.Name), zap.String("MetricType", reflect.TypeOf(metric).String()))
```
NeJan2020, line 93: Use below exp to replace logger.Debug(), see #142
```
if ce := p.telemetry.Logger.Check(zapcore.DebugLevel, "Undefined metricKind for this Metric"); ce != nil {
	ce.Write(zap.String("MetricName", metric.Name), zap.String("MetricType", reflect.TypeOf(metric).String()))
}
```

deploy/agent/kindling-collector-config.yml
NeJan2020: We can also only remove the average time from aggregate_kind_map and keep them in the exporter.metric_aggregation_map.

dxsup: Hmm, it is an option indeed. But we can't control which histogram aggregated in that way.

NeJan2020 (approved):

PR #282, opened 2022-06-30
Declare the 9500 port in the agent's deployment file

dxsup:
This is necessary because we want to display the agent in the topology.

### DIFF
--- a/deploy/agent/kindling-deploy.yml
+++ b/deploy/agent/kindling-deploy.yml
@@ -28,8 +28,6 @@ spec:
             memory: 300Mi
         ports:
         - containerPort: 9500
-          hostPort: 9500
-          name: http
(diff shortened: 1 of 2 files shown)


deploy/agent/kindling-deploy.yml
```diff
@@ -26,6 +26,11 @@ spec:
             memory: 1Gi
           requests:
             memory: 300Mi
+        ports:
+        - containerPort: 9500
+          hostPort: 9500
+          name: http
```
NeJan2020, line 32: Should we decare hostPort when using `hostNetWork` policy

dxsup: No. I'll remove it.

NeJan2020 (approved):

PR #285, opened 2022-06-30
Allow the collector run in the non-Kubernetes environment

dxsup:
## Description

Add a `disable` option, which is `false` by default, to the configuration of the `kubernetesprocessor`. If it is true, the `kubernetesprocessor` won't try to connect to the Kubernetes API-server and will just send the `dataGroups` it consumes to the next consumer.

## Motivation and Context

Some users want to run the collector in the non-Kubernetes environment.

There is also another implementation that meets the requirement. If the `kubernetesprocessor` fails to connect the API-server when initializing, we don't panic the process and just send the data to the next consumer later. In this way, no option is added. But the problem is that the users don't know whether the connection is ok. By explicitly setting the `disable` option, the users know the behavior of the processor. So I prefer panicking the process to letting it run if the connection failed and the option is not set, in which way the users could know exactly what happened.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,7 +5,7 @@
 
 ## Unreleased 
 ### Enhancements
-- Allow the collector run in the non-Kubernetes environment by setting the option `disable` true under the `k8smetadataprocessor` section. ([#285](https://github.com/CloudDectective-Harmonycloud/kindling/pull/285))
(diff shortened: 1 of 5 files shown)


NeJan2020 (approved):
LGTM

NeJan2020 (approved):

PR #287, opened 2022-07-06
fix: correct external topology metric name

NeJan2020:
We used `kindling_entity_request`  to name those request metrics captured from the server-side which shows the working status of the service, the key to identifying server-side metrics is a label named 'is_server'. 

That makes a mistake when we are trying to use server-side metrics to show topology from outside the cluster.

The generated topology should be named with the prefix `kindling_topology_request`, but now they are named with the prefix `kindling_entity_request` since the  'is_server' label is true.

This PR will fix this.


dxsup:
Please update the changelog.

dxsup (approved):

PR #291, opened 2022-07-13
achieve the page fault

yaofighting:
Implement the monitoring of major & minor page fault based on page fault tracepoint
## Description

In the implementation of this version, our events come from 3 places: one is the initialization information of page fault, which we get from /proc, process it as data under thread granularity, and store it in thread table, and then we further encapsulate these threads as kindlingevent to maintain the consistency with the upper collector data interface; The second is from the capture of page fault events by the perf buffer. When the kernel captures a process exit event, it will send data to the user space, and we will consume these data in the collector.The third source is an eBPF Map, which reads the pagefault value in the Map at certain intervals(5 seconds by default) in user mode.
image

## Motivation and Context

We need to use page fault information to judge the memory usage of system processes or threads, so that we can troubleshoot problems and faults. We provide kindling_pagefault_major_total and kindling_pagefault_ minor_total, they have the following meanings:
(1) kindling_pagefault_major_total: Indicates that the system's major page fault data is captured, such as MMAP, SWAP and other scenarios may cause the generation of major page fault, which is associated with a specific thread that generates the page fault. We identify it with 11 labels {node, namespace, workload_kind, workload_name, service, pod, container, container_id, IP, PID, tid}.
(2) kindling_pagefault_minor_total: It indicates that the minor page fault data of the system is captured. For example, using malloc function to apply for and use memory will cause minor page fault, which is also associated with a specific thread that generates the page fault. Similarly, we still use 11 labels {node, namespace, workload_kind, workload_name, service, pod, container, container_id, IP, PID, tid} to identify it. 

## How Has This Been Tested?

Here we use the 'stressapptest' program to increase memory pressure for the system (the tool project address is: （https://github.com/stressapptest/stressapptest. Stressful Application Test (or stressapptest, its unix name) is a memory interface test. It tries to maximize randomized traffic to memory from processor and I/O, with the intent of creating a realistic high load situation in order to test the existing hardware devices in a computer. It has been used at Google for some time and now it is available under the apache 2.0 license. We can use it for correctness verification and performance testing.
We accumulate the thread data belonging to the same streeaptest process, and compare the result with `ps -eo Maj_ flt,min_ The results under FLT, PID, CMD ` commands are compared, so that the correctness can be verified. We use it to put pressure on the memory to test the performance of the program

kernel version: 
Linux localhost.localdomain 3.10.0-1160.el7.x86_64 
Linux version:
CentOS 7.5

The core processing logic of this function is mainly completed in agent-libs. In kindling, we mainly realize the conversion, processing and encapsulation of initialization data and the implementation of page fault pipeline. All work is carried out on the original architecture, but getpagefaultevent is added


dxsup:
Please update the changelog and the metrics description doc.

dxsup (approved):

collector/pkg/component/analyzer/pgftmetricanalyzer/pgft_analyzer.go
```diff
@@ -0,0 +1,144 @@
...
+	containerId := threadinfo.GetContainerId()
+
+	tid := (int64)(threadinfo.GetTid())
+	pid := (int64)(threadinfo.GetPid())
+
+	var localNodeName string
+	var err error
+	if localNodeName, err = getHostNameFromEnv(); err != nil {
+		return labels, fmt.Errorf("localName not found with %s", event.Name)
+	}
```
NeJan2020, line 126: Try to use a static string to save the HostName instead of calling os.GetEnv() every time

collector/pkg/component/consumer/processor/k8sprocessor/kubernetes_processor.go
```diff
@@ -78,12 +78,48 @@ func (p *K8sMetadataProcessor) Consume(dataGroup *model.DataGroup) error {
...
 	return p.nextConsumer.Consume(dataGroup)
 }
 
+var cnt int = 0
+
+func (p *K8sMetadataProcessor) processPgftMetric(dataGroup *model.DataGroup) {
+	if cnt%10 == 0 {
+		p.telemetry.Logger.Info("pagefault_from_processor: ", zap.Int("num:", cnt))
+	}
+	cnt++
```
NeJan2020, line 95: Maybe you can add an extra debug option for this log

collector/pkg/component/receiver/cgoreceiver/cgoreceiver.go
```diff
@@ -63,10 +63,40 @@ func (r *CgoReceiver) Start() error {
...
+		if res == -1 {
+			break
+		}
+		pageCnt++
+		event := convertEvent((*CKindlingEventForGo)(pKindlingEvent))
+		r.eventChannel <- event
+		r.stats.add(event.Name, 1)
+	}
+	r.telemetry.Logger.Info("pagefault init:", zap.Int("the total number of init pagefault threads", pageCnt))
+
```
NeJan2020, line 97: The `zap.Int(key,value)`  is used to export the struct as a JSON object.
Please move the error msg to the first param of `logger.Info`,e.g
```go
r.telemetry.Logger.Info("pagefault init successed",zap.Int("threads_count",pageCnt))
```

collector/pkg/component/analyzer/pgftmetricanalyzer/pgft_analyzer.go
yaofighting: OK, I put the method of getting localname into newPgftMetricAnanalyzer().

NeJan2020 (dismissed):

probe/src/cgo/kindling.cpp
```diff
@@ -133,6 +139,95 @@ void init_probe()
...
+	strcpy(p_kindling_event->context.tinfo.comm, (char *)threadInfo->m_comm.data());
+	strcpy(p_kindling_event->context.tinfo.containerId, (char *)threadInfo->m_container_id.data());
+	strcpy(p_kindling_event->name, "page_fault");
+
+	return 1;
+}
+
+void initPageFaultOffData(){
+	threadinfo_map_t *threadsmap = inspector->m_thread_manager->get_threads();
+    threadstable = threadsmap->getThreadsTable();
```
jundizhou, line 221: Please note the code format

NeJan2020:
This feature will make the Kindling-agent crash when the kernel doesn't fully support eBPF, such as CentOS 7.8 (3.10.0-1127) or below.
Can you add more meaningful error messages when the program subscribes to those unsupported events?

dxsup:
Also, we need to make sure the users know what features are supported under which kernel versions. Knowing only after failure really harms users' experience. I think a document `Supported Feature List` can make it.

> This feature will make the Kindling-agent crash when the kernel doesn't fully support eBPF, such as CentOS 7.8 (3.10.0-1127) or below.
> Can you add more meaningful error messages when the program subscribes to those unsupported events?

dxsup:
Must we always add hundreds of lines of code to `kindling.cpp`? There have been more than one thousand lines in this file.  Can you move the page-fault related code out of it and give the new file a meaningful name?

collector/docker/kindling-collector-config.yml
```diff
@@ -118,6 +122,10 @@ processors:
         - kind: sum
       response_io:
         - kind: sum
+      kindling_pagefault_major_total:
+        - kind: last
+      kindling_pagefault_minor_total:
+        - kind: last
```
dxsup, line 128: If the type of the two metrics is "last", then the metric name should not be ended with "_total" which is designed for the "counter" type.

Issue #298, opened 2022-07-28
Need to improve log format


NeJan2020:
There are some problems with our application log.

1. JsonString or detail string should be printed as a message, not a field of `zap.logger`

Now:
https://github.com/CloudDectective-Harmonycloud/kindling/blob/650f8b6cd7c50dd37b69bbad2dfb987668ed0619/collector/pkg/component/receiver/cgoreceiver/cgoreceiver.go#L154-L158

Changed into:
```go
if ce := r.telemetry.Logger.Check(zapcore.DebugLevel, "Receive Event"); ce != nil { 
 	 r.telemetry.Logger.Debug("EventDetail: "+evt.String())
 } 
```
This can prevent JSON strings from being repeatedly escaped.

Output Now:
```text
[...]

PR #299, opened 2022-07-28
enhance/improve log format

NeJan2020:
## Description
- overwrite MarshalJSON() and UnmarshalJSON() for attributeMap for those private Members
- move those outputs in JSON from `zap.filed` to message  

## Related Issue
#298 

## After Change
KindlingEvent
```text
2022-07-28T18:35:45.733Z        DEBUG   cgoreceiver/cgoreceiver.go:156  Receive Event: &{Source:SOURCE_UNKNOWN Timestamp:1659033341930367401 Name:connect Category:3 ParamsNumber:3 UserAttributes:[latency:UINT(306428) res:INT(0) tuple:BYTEBUF(^A^Y^^@^Y^/var/run/docker.sock^@)] Ctx:{ThreadInfo:{Pid:11474 Tid:11608 Uid:0 Gid:0 Comm:kubelet ContainerId: ContainerName:} FdInfo:{Num:10 TypeFd:8 Filename: Directory: Protocol:0 Role:false Sip:0.0.0.0 Dip:0.0.0.0 Sport:0 Dport:0 Source:18446622435791445440 Destination:18446622435791437824}}}
```
DataGroup
```text
2022-07-28T18:35:45.741Z        DEBUG   network/network_analyzer.go:322 NetworkAnalyzer To NextProcess:
DataGroup:
        Name: net_request_metric_group
        Values:
                "connect_time": 0
                "request_sent_time": 21010
                "waiting_ttfb_time": 1197746594
                "content_download_time": 656720
                "request_total_time": 1198424324
                "request_io": 35
                "response_io": 10564
        Labels:
        {
                "comm": "kindling-collec",
                "container_id": "c4ffdfa6ebfa",
                "dnat_ip": "10.10.101.124",
                "dnat_port": 6443,
                "dst_ip": "10.96.0.1",
                "dst_port": 443,
                "error_type": 0,
                "is_error": false,
                "is_server": false,
                "is_slow": true,
                "pid": 32433,
                "protocol": "NOSUPPORT",
                "src_ip": "10.10.103.159",
                "src_port": 58178
        }
        Timestamp: 1659033342510904252
```

### DIFF
--- a/collector/pkg/model/attribute_map.go
+++ b/collector/pkg/model/attribute_map.go
@@ -17,6 +17,14 @@ type AttributeMap struct {
 	values map[string]AttributeValue
 }
 
+func (a AttributeMap) MarshalJSON() ([]byte, error) {
+	return json.Marshal(a.values)
+}
+
+func (a AttributeMap) UnmarshalJSON(data []byte) error {
+	return json.Unmarshal(data, &a.values)
+}
+
 func NewAttributeMap() *AttributeMap {
 	values := make(map[string]AttributeValue)
 	return &AttributeMap{values}
@@ -175,6 +183,14 @@ func (v *stringValue) Type() AttributeValueType {
 	return StringAttributeValueType
 }
 
+func (v stringValue) MarshalJSON() ([]byte, error) {
+	return json.Marshal(v.value)
+}
+
+func (v *stringValue) UnmarshalJSON(data []byte) error {
+	return json.Unmarshal(data, &v.value)
+}
+
 func (v *stringValue) ToString() string {
 	return v.value
 }
@@ -195,6 +211,14 @@ func (v *intValue) ToString() string {
 	return strconv.FormatInt(v.value, 10)
 }
 
+func (v intValue) MarshalJSON() ([]byte, error) {
+	return json.Marshal(v.value)
+}
+
+func (v *intValue) UnmarshalJSON(data []byte) error {
+	return json.Unmarshal(data, &v.value)
+}
+
 func (v *intValue) Reset() {
 	v.value = 0
 }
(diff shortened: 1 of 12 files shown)


collector/pkg/model/attribute_map.go
```diff
@@ -17,6 +17,14 @@ type AttributeMap struct {
 	values map[string]AttributeValue
 }
 
+func (a AttributeMap) MarshalJSON() ([]byte, error) {
+	return json.Marshal(a.values)
+}
+
+func (a AttributeMap) UnmarshalJSON(data []byte) error {
+	return json.Unmarshal(data, &a.values)
+}
```
dxsup, line 26: Why not use the pointer as the function receiver?

collector/pkg/model/attribute_map.go
```diff
@@ -175,6 +183,14 @@ func (v *stringValue) Type() AttributeValueType {
 	return StringAttributeValueType
 }
 
+func (v stringValue) MarshalJSON() ([]byte, error) {
```
dxsup, line 186: Why not use the pointer as the function receiver?

collector/pkg/model/attribute_map.go
NeJan2020: In example of [encoding/json](https://pkg.go.dev/encoding/json) ,  `MarshalJSON()` should use the obj as the receiver ,and `UnmarshalJSON(b []byte)` should use the pointer as the receiver . 

We usually call `json.Unmarshal([]bytes, &obj)` and `json.Marshal(obj)` to invoke functions above ,I guess that's the reason. I will change the receiver of `UnmarshalJSON()`'

``` go
func (a *Animal) UnmarshalJSON(b []byte) error {
	...
}

func (a Animal) MarshalJSON() ([]byte, error) {
	...
}
```

collector/pkg/model/attribute_map.go
NeJan2020: Same as above

dxsup (approved):

PR #315, opened 2022-09-16
exit when failed to start probe

sanyangji:
## Motivation
I notice that if probe fails to start, huge logs of program stacks will appear because of nil pointer, which makes no sense.
Just exit program when failed to start probe, make the error log obvious.

No image here.


dxsup:
I think we better move the exit to the Go part. This can make its behavior more consistent with the current one.

dxsup (approved):

Issue #366, opened 2022-11-23
Add the observation function for slow, timeout and error system calls


yaofighting:
**Background**
With the increase of the number of applications and services in the system, we often need to locate some system calls that run for a long time or have errors. They are often bottlenecks that cause service or application problems. 

**Solution**
System call information is obtained in the kernel based on eBPF technology. System calls are analyzed, processed and correlated in user space. Long running system calls are divided into slow system calls and timeout system calls. Slow system calls will be matched immediately according to _enter_ and _exit_, and overtime system calls will be judged by the timed polling mechanism.The timeout time of the timeout system call and the slow threshold determination time of the slow system call are both configured by the user. 

**Features**
[1] Modify the configuration file structure and add parameter fields for subscription events. 
[2] Achieve the Slow Syscall function. 
[3] Achieve the Timeout Syscall function.

dxsup:
There is still a lot of work to do. Expect this can be finished before February.

dxsup:
We won't be scheduling this feature in the near future since we haven't identified a valuable use case for it yet. We need more information before we can decide whether to continue working on it.

PR #368, opened 2022-11-23
Modify the configuration file structure and add parameter fields for subscription events

yaofighting:
Modify the configuration file structure and add parameter fields for subscription events

### Motivation
In some scenarios, only specifying the type of subscription event cannot meet our needs. Therefore, we need to enhance the features of the subscription part, support more detailed parameter lists, and control the CPP side to send only the event information we are concerned about to the Go side. For example, we need to specify system calls that filter latency is less than 500ms. When subscribing to events, we tell the CPP side to send only data that meets this condition, so as to avoid performance loss caused by data filtering on the Go side.
## Related Issue
#366

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - 
 - Support the protocol RocketMQ.([#328](https://github.com/KindlingProject/kindling/pull/328))
 - Add a new tool: A debug tool for Trace Profiling is provided for developers to troubleshoot problems.([#363](https://github.com/CloudDectective-Harmonycloud/kindling/pull/363))
+- Modify the configuration file structure and add parameter fields for subscription events. ([#368](https://github.com/CloudDectective-Harmonycloud/kindling/pull/368))
 
 
 ### Enhancements
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,14 +5,12 @@
 
 ## Unreleased
 ### New features
-- 
 - Support the protocol RocketMQ.([#328](https://github.com/KindlingProject/kindling/pull/328))
 - Add a new tool: A debug tool for Trace Profiling is provided for developers to troubleshoot problems.([#363](https://github.com/CloudDectective-Harmonycloud/kindling/pull/363))
-- Modify the configuration file structure and add parameter fields for subscription events. ([#368](https://github.com/CloudDectective-Harmonycloud/kindling/pull/368))
 
 
 ### Enhancements
-- 
+- Modify the configuration file structure and add parameter fields for subscription events. ([#368](https://github.com/CloudDectective-Harmonycloud/kindling/pull/368))
 - 
 - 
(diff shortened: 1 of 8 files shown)


sanyangji (dismissed):
LGTM.
We may unify the Naming Convention for cpp codes and cgo codes.
 @dxsup @jundizhou

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
 - 
 - Support the protocol RocketMQ.([#328](https://github.com/KindlingProject/kindling/pull/328))
 - Add a new tool: A debug tool for Trace Profiling is provided for developers to troubleshoot problems.([#363](https://github.com/CloudDectective-Harmonycloud/kindling/pull/363))
+- Modify the configuration file structure and add parameter fields for subscription events. ([#368](https://github.com/CloudDectective-Harmonycloud/kindling/pull/368))
```
dxsup, line 11: I consider it more like an "Enhancement" than a "New feature".

dxsup:
> LGTM.
> We may unify the Naming Convention for cpp codes and cgo codes.
> @dxsup @jundizhou

For example? I didn't see any difference on "Naming Convention" between these files. Do you mean the indentation style?

CHANGELOG.md
yaofighting: I think so,  I will modify it immediately.

sanyangji:
> > LGTM.
> > We may unify the Naming Convention for cpp codes and cgo codes.
> > @dxsup @jundizhou
> 
> For example? I didn't see any difference on "Naming Convention" between these files. Do you mean the indentation style?

``

> > LGTM.
> > We may unify the Naming Convention for cpp codes and cgo codes.
> > @dxsup @jundizhou
> 
> For example? I didn't see any difference on "Naming Convention" between these files. Do you mean the indentation style?
`void sub_event(char* eventName, char* category, event_params_for_subscribe params[])`

sanyangji:
> > LGTM.
> > We may unify the Naming Convention for cpp codes and cgo codes.
> > @dxsup @jundizhou
> 
> For example? I didn't see any difference on "Naming Convention" between these files. Do you mean the indentation style?

`void sub_event(char* eventName, char* category, event_params_for_subscribe params[])` Like In this line, eventName is different from others

dxsup:
OK, so first let's determine the convention on naming, then we can rename them all.

sanyangji (approved):

dxsup:
We can open another PR for the "naming convention". Back to this one, LGTM.

dxsup (approved):

yaofighting:
I think we should use _underscore_case_ in C/C++ and _camelCase_ in Golang.

yaofighting:
I think we should use _underscore_case_ in C/C++ and _camelCase_ in Golang.

dxsup (approved):

PR #376, opened 2022-11-29
Add no_response_threshold(120s) for No response requests

hocktea214:
Signed-off-by: huxiangyuan <huxiangyuan@harmonycloud.cn>

## Description

There're two checks for no response request and completed request with same 15s check.
In fact, complete request should check timely with 15s
But the no response request should wait long period with 120s which is set with no_response_threshold in config file.

Besides, there may be multi-send with same request, use atomic  check for send once.

### DIFF
--- a/collector/docker/kindling-collector-config.yml
+++ b/collector/docker/kindling-collector-config.yml
@@ -42,8 +42,10 @@ analyzers:
   tcpmetricanalyzer:
   networkanalyzer:
     connect_timeout: 100
+    # How many seconds to wait until we consider a request as complete.
+    request_timeout: 15
     # How many seconds to wait until we consider a request as no response.
-    request_timeout: 60
+    no_response_threshold: 120
     # How many milliseconds to wait until we consider a request-response as slow.
     response_slow_threshold: 500
     # Whether enable conntrack module to find pod's ip when calling service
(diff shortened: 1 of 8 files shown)


collector/docker/kindling-collector-config.yml
```diff
@@ -42,8 +42,10 @@ analyzers:
   tcpmetricanalyzer:
   networkanalyzer:
     connect_timeout: 100
+    # How many seconds to wait until we consider a request as complete.
+    request_timeout: 15
```
dxsup, line 46: I think "request_timeout" is ambiguous now because it is used to check whether `fds` is reused. Better change to `fd_reuse_timeout`.

collector/pkg/component/analyzer/network/network_analyzer.go
```diff
@@ -193,8 +193,15 @@ func (na *NetworkAnalyzer) consumerFdNoReusingTrace() {
...
 				var timeoutTs = mps.getTimeoutTs()
-				if timeoutTs != 0 && (time.Now().UnixNano()/1000000000-int64(timeoutTs)/1000000000) >= 15 {
-					na.distributeTraceMetric(mps, nil)
+				if timeoutTs != 0 {
+					var duration = (time.Now().UnixNano()/1000000000 - int64(timeoutTs)/1000000000)
+					if mps.responses != nil && duration >= int64(na.cfg.GetRequestTimeout()) {
+						// No New Request
+						na.distributeTraceMetric(mps, nil)
+					} else if duration >= int64(na.cfg.getNoResponseThreshold()) {
+						// Request 498
```
dxsup, line 202: "498" is ambiguous. How about changing it to "No response"?

dxsup (approved):

PR #381, opened 2022-12-05
Add response variable name for ParsePkgFn

blue-troy:
Add response variable name for ParsePkgFn(#382)

Signed-off-by: blue-troy <12729455+blue-troy@users.noreply.github.com>


dxsup:
Seems like this PR has not been finished yet and I can't get why this PR is necessary. Could you please describe the motivation for what you want to change? Maybe you should open an issue so that we can talk about it.

blue-troy:
#382

dxsup (approved):

hocktea214:
It seems as a case for evolution of function.

hocktea214 (approved):

PR #385, opened 2022-12-07
Fix the wrong thread name in the trace profiling function.

yaofighting:

Fix the thread name error in the trace profiling function, and optimize the segment field in the cycle_queue in collector. 
### Analysis-Probe
When the java probe thread itself generates its own java_futex information for the first time, it enters Probe-side ` parse_ jf() ` function, at this time ` parse_ tm() ` thread has not been processed yet, so the ` ptid_comm` map has no thread name information of this thread. However, in the code logic, it only writes the logic to assign the thread name when the thread exists in the `ptid_comm`map. When the map does not exist, it does nothing. (Therefore, the comm field in kindling_event is actually the comm of the previous data.) like this: 
`if (key != ptid_comm.end()) {
    strcpy(p_kindling_event->context.tinfo.comm, key->second);
  }`
### Analysis-Collector
At this time, the incorrect comm data will enter the `ConsumeEvent ->ConsumeJavaFutexEvent ->PutEventToSegments` logic in the Go side cpuanalyzer to establish the first entry information of the thread. At this time, a ring will be initialized and the thread name information will be initialized. Unless the ring is cleaned up later, the thread name information will not be updated in a timely manner, so it's also a problem in collector. 
### Solution
For theProbe-side, add the default comm data value, and for the Collector-side, ensure that the thread name information is updated every time the ring is updated. 

## Problem2. Data redundancy in Collector-Segment
optimize the segment field in the cycle_queue in collector.

### DIFF
--- a/collector/pkg/component/analyzer/cpuanalyzer/cpu_analyzer.go
+++ b/collector/pkg/component/analyzer/cpuanalyzer/cpu_analyzer.go
@@ -218,10 +218,11 @@ func (ca *CpuAnalyzer) PutEventToSegments(pid uint32, tid uint32, threadName str
 
 	} else {
 		newTimeSegments := &TimeSegments{
-			Pid:      pid,
-			Tid:      tid,
-			BaseTime: event.StartTimestamp() / nanoToSeconds,
-			Segments: NewCircleQueue(maxSegmentSize),
+			Pid:        pid,
+			Tid:        tid,
+			ThreadName: threadName,
+			BaseTime:   event.StartTimestamp() / nanoToSeconds,
+			Segments:   NewCircleQueue(maxSegmentSize),
 		}
 		for i := 0; i < maxSegmentSize; i++ {
 			segment := newSegment((newTimeSegments.BaseTime+uint64(i))*nanoToSeconds,
--- a/collector/pkg/component/analyzer/cpuanalyzer/cpu_analyzer.go
+++ b/collector/pkg/component/analyzer/cpuanalyzer/cpu_analyzer.go
@@ -181,8 +181,7 @@ func (ca *CpuAnalyzer) PutEventToSegments(pid uint32, tid uint32, threadName str
 				endOffset = endOffset - startOffset
 				startOffset = 0
 				for i := 0; i < maxSegmentSize; i++ {
-					segment := newSegment(pid, tid, threadName,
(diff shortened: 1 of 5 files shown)


probe/src/cgo/kindling.cpp
```diff
@@ -430,6 +430,8 @@ void parse_jf(char* data_val, sinsp_evt_param data_param, kindling_event_t_for_g
       ptid_comm.find(threadInfo->m_pid << 32 | (threadInfo->m_tid & 0xFFFFFFFF));
   if (key != ptid_comm.end()) {
     strcpy(p_kindling_event->context.tinfo.comm, key->second);
+  }else{
```
dxsup, line 433: Please reformat the code.

collector/pkg/component/analyzer/cpuanalyzer/model.go
```diff
@@ -61,6 +57,9 @@ func newSegment(pid uint32, tid uint32, threadName string, startTime uint64, end
 		IndexTimestamp:  "",
 	}
 }
+func (t *TimeSegments) updateThreadName(threadName string) {
+	t.ThreadName = threadName
+}
```
dxsup, line 62: Better put this method under the definition of its receiver `TimeSegments` to improve the readability.

yaofighting: ok， I have formatted the code.

collector/pkg/component/analyzer/cpuanalyzer/cpu_analyzer.go
```diff
@@ -225,8 +224,7 @@ func (ca *CpuAnalyzer) PutEventToSegments(pid uint32, tid uint32, threadName str
 			Segments: NewCircleQueue(maxSegmentSize),
 		}
 		for i := 0; i < maxSegmentSize; i++ {
-			segment := newSegment(pid, tid, threadName,
-				(newTimeSegments.BaseTime+uint64(i))*nanoToSeconds,
+			segment := newSegment((newTimeSegments.BaseTime+uint64(i))*nanoToSeconds,
```
dxsup, line 227: Did you forget to initialize the field `ThreadName` of `TimeSegments`?

dxsup (approved):

PR #398, opened 2022-12-15
Add events statistics log and gdb information when unexpected exit happens

jundizhou:
…f exit


dxsup (dismissed):

probe/src/cgo/catch_sig.cpp
```diff
@@ -0,0 +1,74 @@
...
+  sprintf(cmd, "gcore %u", getpid());
+  system(cmd);
+  char path[PATH_MAX];
+  char processname[1024];
+  printf("get pid file.\n");
+  fflush(stdout);
+  get_executable_path(path, processname, sizeof(path));
+  // TODO!!!!
+  sprintf(cmd, "./pl/gdb_print.sh ./pl/%s ./core.%u", processname, getpid());
+  system(cmd);
```
sanyangji, line 45: Looks like miss gdb_print.sh.
And changing the script path is better

dxsup (approved):

Issue #401, opened 2022-12-16
network analyzer nil point


blue-troy:
**Describe the bug**
network analyzer nil point

**Screenshots**

**What config did you use?**
Config: (e.g. the yaml config file)

**Logs**
Please attach the logs by running the following command:
```
kubectl logs -f kindling-agent-xxx(replace with your podname) -n kindling -c kindling-collector

2022-12-16T08:42:08.120656454+08:00 stderr F panic: runtime error: invalid memory address or nil pointer dereference
2022-12-16T08:42:08.120709105+08:00 stderr F [signal SIGSEGV: segmentation violation code=0x1 addr=0x8 pc=0xaee839]
2022-12-16T08:42:08.120720095+08:00 stderr F
2022-12-16T08:42:08.120729087+08:00 stderr F goroutine 128 [running]:
2022-12-16T08:42:08.120758518+08:00 stderr F github.com/Kindling-project/kindling/collector/pkg/component/analyzer/network.(*NetworkAnalyzer).getRecordWithSinglePair(0xc0003aa420
, 0xc001ff9b00, 0xc0023736c8, {0x1b134d3, 0x0}, 0xc)
2022-12-16T08:42:08.120781491+08:00 stderr F    /kindling/collector/pkg/component/analyzer/network/network_analyzer.go:616 +0xe99
2022-12-16T08:42:08.120791015+08:00 stderr F github.com/Kindling-project/kindling/collector/pkg/component/analyzer/network.(*NetworkAnalyzer).parseMultipleRequests(0xc002373848,
0xc001ff9b00, 0xc00103e480)
2022-12-16T08:42:08.120799386+08:00 stderr F    /kindling/collector/pkg/component/analyzer/network/network_analyzer.go:495 +0x891
2022-12-16T08:42:08.120816296+08:00 stderr F github.com/Kindling-project/kindling/collector/pkg/component/analyzer/network.(*NetworkAnalyzer).parseProtocol(0x187d100, 0xc002373a0
0, 0xc00054fda0)
2022-12-16T08:42:08.120827133+08:00 stderr F    /kindling/collector/pkg/component/analyzer/network/network_analyzer.go:413 +0x40b
2022-12-16T08:42:08.120853127+08:00 stderr F github.com/Kindling-project/kindling/collector/pkg/component/analyzer/network.(*NetworkAnalyzer).parseProtocols(0xc0003aa420, 0xc001f
f9b00)
[...]

PR #402, opened 2022-12-16
Fix network analyzer nil pointer

blue-troy:
Fix network analyzer nil point

Signed-off-by: blue-troy <12729455+blue-troy@users.noreply.github.com>


blue-troy:
fix #401

dxsup:
Thanks for providing your solution to fix this issue. I have reviewed the original code and found the root cause of why the issue happened is that the method `getRecordWithSinglePair` uses two similar names `mps` and `mp` as parameters. This could cause developers easier to make more mistakes.

We can approve this PR and open a new one to refactor this method.

dxsup (approved):

PR #440, opened 2023-02-02
Feature/soft build

juewu4072:
## Description

This is a commit about the compilation process.
1. Add submodule, and introduce agent-libs as git-submodules to improve locally compilation efficiency.
2. Add make command line to facilitate compilation.

## Motivation and Context

Each time you compile and include the probe locally, you need to extract agent-libs from github twice. The version mapping between agent-libs and kindling is not as clear as the mapping between git submodules. After the git submodule is introduced, git will record the version correspondence. The compilation process contains too many compile commands, which is prone to errors. It is best to use makefiles directly

## How Has This Been Tested?

use command to build
```
$cd kindling/
$make docker-build-all
```
After successful compilation
```
$docker images
```

image


sanyangji:
Good Idea.
I follow your test but find some problems in Makefile. I fail to build the image, can you do more tests and fix them? 
Besides, some new updates should be merged into Dockerfile and take care of typos and code indentation.
One more thing, I'm not sure whether you should take compiling kindling-falcolib-probe tarball into account, as in Makefile, `make probe-libs first`, which will make the compilation process complete.

juewu4072:
sorry to change that bug, the wrong name for the build-image has been fix, and the docker-build-all command was retested successfully.

The dockerfile was modified because it took too much time to install debugging tools such as gdb when packaging the image. we can give them more choices, such as the two common scenarios supported:
1. Need debug tools, need local probe 
2. No debugging tools required, need compile probes

`make probe-libs first` is a good idea, it works well when my computer kernel version is the same, I also use it, since my friend and I compiled with source code on our own computer, because our kernel-version is different, we need to execute make probe-libs every time, to simplify this process, I added several different commands in the makefile for users to choose, for example, 
1. `make dependencies` to install deps
2. `make agent-libs` to build probe-libs
3. `make kindling-collector` to  build collector image
4. `make docker-build-all` to install deps and build probe-libs and build cooector image
 
I added new comments to makefile. @sanyangji

Makefile
```diff
@@ -13,4 +13,135 @@ collector: libkindling
...
+		cp -f kindling-falcolib-probe.tar.gz $(COLLECTOR_PATH)/docker/.; \
+	fi
+
+## build kindling-falcolib-probe in docker
+.PHONY: agent-libs
+agent-libs: build-libs pack-libs
+
+.PHONY: build-collector
+build-collector:
+	@cd $(SCRIPTS_PATH) && sh ./build.sh Dokcerfile
```
sanyangji, line 133: typo: Dockerfile

collector/docker/Dockerfile
```diff
@@ -1,14 +1,13 @@
 FROM centos:7
 
 WORKDIR /app/
-RUN yum install -y qt5-qtbase-devel
-RUN yum install -y gdb
-COPY gdb_print.sh /app/
```
sanyangji, line 6: these are used to load kindling plugin and to export information when unexpected exit happens https://github.com/KindlingProject/kindling/pull/398

Makefile
```diff
@@ -13,4 +13,135 @@ collector: libkindling
...
+
+AGENT_LIBS_COMMIT_ID=$(shell cd $(LIBS_SRC) && git rev-parse HEAD)
+
+.PHONY: patch-libs
+patch-libs:
+	@if [ ! -d "$(LIBS_SRC)" ]; then \
+  		echo "$(LIBS_SRC) not exist, download....."; \
+		git submodule update --init --recursive; \
+	else \
+	  echo "agent-libs: $(LIBS_SRC) exist"; \
```
sanyangji, line 87: inconsistent code indent

Makefile
```diff
@@ -13,4 +13,135 @@ collector: libkindling
...
+		-v $(LIBS_SRC):/source \
+		$(LIBS_IMAGE);
+
+.PHONY: pack-libs
+pack-libs:
+	@echo "Packaging kindling libs...";
+	@if [ ! -d "$(LIBS_SRC)/kindling-falcolib-probe/" ]; then \
+    		echo "The packaged probe does not exist.try again"; \
+    		exit 1; \
+    else \
```
sanyangji, line 121: inconsistent

sanyangji:
Sorry, I misunderstood the meaning of `agent-libs`. `agent-libs` is composed of userspace part (static libs, you may call libs or agent-libs) and kernel space part (that is kindling-falcolib-probe, you may call it  kindling-falcolib-probe or probe but without libs). So can you rename them to make them clear? Thanks to your comments anyway.

Some comments have been added and I'm still trying to test, please wait a moment.

juewu4072:
I've fixed typos, bad naming, code indentation.

my idea has some problems with indent and tab  T_T

.gitmodules
```diff
@@ -0,0 +1,3 @@
+[submodule "agent-libs"]
+	path = agent-libs
```
dxsup, line 2: Since `agent-libs` is one of the dependencies of `probe`, I think we better put the submodule under `kindling/probe`. What do you think?

.gitmodules
```diff
@@ -0,0 +1,3 @@
...
+	url = https://github.com/KindlingProject/agent-libs.git
```
dxsup, line 3: I suggest we write down the branch name explicitly to make things clear.

deploy/scripts/build.sh
```diff
@@ -1,22 +1,33 @@
...
 libKindlingPath="./src/libkindling.so"
 if [ ! -f "$libKindlingPath" ]; then
   echo "compiler libkindling failed! exit!"
 
 else
-  mkdir -p ../../collector/docker/libso &&  cp -rf ./src/libkindling.so ../../collector/docker/libso/
+  mkdir -p ../../collector/docker/libso && cp -rf ./src/libkindling.so ../../collector/docker/libso/
   cp -rf ./src/libkindling.so /usr/lib64/
   cd ../../collector/
+  go mod tidy && go mod vendor
```
dxsup, line 19: Is `go mod vendor` necessary?

dxsup:
Are you still working on this PR? Hope for your response.

.gitmodules
juewu4072: Branching of submodules is allowed, but this is inconsistent with the original design of the submodules, and submodules with specific commits can be used with the current code. Same as libbpf/libbpf bootstrap

juewu4072: `https://github.com/libbpf/libbpf-bootstrap/blob/master/.gitmodules`

deploy/scripts/build.sh
juewu4072: Has been removed in the current version, this will slow down the image packaging speed

.gitmodules
dxsup: `git submodule` does record a specific commit to allow users to pull the correct repository version. There is no question that the current `.gitmodules` works. AFAIK, tracking a remote branch can make things clear because we can use `git submodule update --remote` to update the code. See [How can I specify a branch/tag when adding a Git submodule?](https://stackoverflow.com/questions/1777854/how-can-i-specify-a-branch-tag-when-adding-a-git-submodule).

It is OK if we use the current `.gitmodules` and make an improvement later.

juewu4072: Thank you for your help. That's a great idea. Adding submodule branches is a great way to improve multiplayer collaboration, and it solves a problem I had with using submodule multiplayer collaboration scenarios. Thanks again

.gitmodules
juewu4072: The agent library is a dependency of the probe, and I think it makes sense to do that. I'll put it under the kindling/probe/libs/.

deploy/scripts/build.sh
dxsup: AFAIK, `go mod vendor` is helpful in two cases:
1. When there is no Internet access.
2. When some packages are not available from the proxy anymore.

Kindling has been using `GOPATH` as the pkg directory and it works well. I have not seen the need to use `go mod vendor`.

Regarding the image packaging speed, unless we maintain the `vendor` directory in the Git repository, there is no noticeable difference, IMO.

juewu4072: When we compile  `GOPATH = /root/go/` ，use run_docker. sh again, path ‘/root/go’ does not exist, need to download modules again. If we do vendors, the vendor path  `/kindling/collector/vendor/` will be mapped to images with Kindling code. Compiling again will reduce the time. This is only valid for multiple partial compilations. 

It's a great help to compile multiple times when we're local.

juewu4072: This scenario is friendly for local build testing, but has some concerns about being a generic scenario, so I've removed it in a new commit, which should be done by tomorrow evening at the latest

.gitmodules
juewu4072: agent-libs has moved to 'kindling/probe/libs/agent-libs' and add branch mark in .gitmodules

deploy/scripts/build.sh
juewu4072: removed 'go mod vendor'

dxsup:
Thanks for your work again. I'll have a test after which I'll merge this PR.

dxsup:
@sanyangji @jundizhou Do we still need to maintain the commit hash of `agent-libs` in `CMakeLists.txt` since this PR has introduced `git submodule`?

sanyangji:
> @sanyangji @jundizhou Do we still need to maintain the commit hash of `agent-libs` in `CMakeLists.txt` since this PR has introduced `git submodule`?
I have no idea about the usage of the `commit` if merging this PR. It's better to remove it.

Makefile
```diff
@@ -13,4 +13,144 @@ collector: libkindling
...
+
+.PHONY: pack-probe
+pack-probe:
+	@echo "Packaging kindling probe...";
+	@if [ ! -d "$(LIBS_SRC)/kindling-falcolib-probe/" ]; then \
+		echo "The packaged probe does not exist.try again"; \
+		exit 1; \
+	else \
+		cd $(LIBS_SRC) && \
+		tar -cvzf kindling-falcolib-probe.tar.gz kindling-falcolib-probe/ && \
```
NeJan2020, line 132: Please add the `probe/libs/agent-libs/kindling-falcolib-probe.tar.gz` to the .gitignore of `agent-lib`, or just remove it after copy.

juewu4072: Changed `cp -f` to `mv`

juewu4072: Have also added `clean` to the makfile, clean the build file

juewu4072:
> 

Removed useless patch

Makefile
```diff
@@ -13,4 +13,149 @@ collector: libkindling
...
+		echo "$(KernelName) already installed"; \
+	fi
+
+## Check libs exists or download, patch cmake file
+.PHONY: patch-libs
+patch-libs:
+	@echo "Checking libs...";
+	@if [ ! -d "$(LIBS_SRC)" ]; then \
+		echo "$(LIBS_SRC) not exist, downloading....."; \
+		git submodule update --init --recursive; \
```
dxsup, line 89: The `if` statement is always false because `agent-libs` is a directory and it's always there once you clone a fresh Kindling repository.

Makefile
```diff
@@ -13,4 +13,149 @@ collector: libkindling
...
+	ifneq ($(shell cat /etc/os-release | grep CentOS | wc -l),0)
+		KernelName = kernel-devel-$(kernelVersion)
+		CheckCommand = yum list installed |grep $(KernelName)
+		InstallCommand = sudo yum -y install $(KernelName)
+	else
+		KernelName = NotSupport
+		CheckCommand = $(ErrorExit)
+		InstallCommand = $(ErrorExit)
+	endif
+endif
```
dxsup, line 43: Looks like these are only needed when running `make kernel-headers`. Why not move these under that part?

By the way, I figured `CheckCommand` is never used and should be removed.

Makefile
juewu4072: Yes, your findings are valuable, and that really shouldn't be judged by `if -d $path`.

juewu4072: Changed to `ls -A1q $path` to check dir is empty

Makefile
juewu4072: Moved to `make kernel-headers`.

dxsup (approved):
LGTM

dxsup:
I'm writing down TODOs here once we merge this PR:
- [ ] Fix the action `Kindling-agent-CI`.
- [ ] Rewrite the doc guiding developers on how to build.

PR #455, opened 2023-02-17
Fix improve cgoreceiver log format

NeJan2020:
Optimized the log output of `CgoReceiver` to display the specific value in UserAttribute

## Description
Print Specific Value of UserAttributes rather than base64 Encoding.

Only print printable character (ASCII char code from 32 to 126) for ByteBuffer , other character will be replaced by `.`
The max length is 200, and extra bytes will be count, such as `<POST /test?sleep=800 HTTP/1.1xxxxx(69 bytes more)>`

## How Has This Been Tested?
```
Running tool: /usr/local/go/bin/go test -timeout 60m -run ^TestTextEvent$ github.com/Kindling-project/kindling/collector/pkg/model

=== RUN   TestTextEventa
--- PASS: TestTextEvent (0.00s)
PASS
ok      github.com/Kindling-project/kindling/collector/pkg/model        0.010s
```

Here is some examples.
```
{Name:read Category:3 ParamsNumber:3 UserAttributes:[starttime:1666085694803271345 data:<POST /test?sleep=800 HTTP/1.1...Host: 10.10.103.148:9000...Content-Type: application/x-www-form-urlencoded ...xxxxjkajdnaksndandlkajelkqjweklqwjeklqjlwejklasdas...qwelkqnmwlkenqlwenqwnekqkwelkqmnwlekq(69 bytes more)>] Ctx:{ThreadInfo:{Pid:22592 Tid:22592 Uid:0 Gid:0 Comm:node ContainerId:123123 ContainerName:test} FdInfo:{Num:19 TypeFd:3 Filename:tmp Directory:/root Protocol:1 Role:true Sip:[16777343] Dip:[16777343] Sport:39620 Dport:37093 Source:0 Destination:0}} Latency:100 Timestamp:160000}
```


dxsup:
> Only print char in [a-zA-Z0-9] and `.` `/` `\` `@` `?` `-` `=`   for ByteBuffer now, other char will be replaced by ` `,

I think `.` is better because Sysdig does like it which we have been familiar with.

> and the max length is 80

80 is not enough since we set `SNAPLEN` to 1000 by default. We better set the max length to 200 at least and add a sign if there are more than 200 bytes to explicitly tell users we truncate the bytes - maybe we could add an `xxx bytes more` sign at the tail.

dxsup (dismissed):

dxsup (approved):

PR #516, opened 2023-05-13
Add label select for k8s metadata processor

llhhbc:
## Description

Add label selector for k8s metadata processor.
For only need watch some pod by label selector.
I'm not sure if I should limit which pods to fetch only in origin.


dxsup:
Could you please also provide an example of the configuration for selecting specified pods? It seems like you are using a set of "pod's labels" to match the "DataGroup's labels" which contain no k8s metadata when consumed by this processor.

llhhbc:
I just ues the struct of kubernetes, It's configable. I will add some examples.

llhhbc:
@dxsup Can this be merged?

dxsup:
Sorry for the late reply. I guess the intention of this PR is to filter the `DataGroups` that match the labels given in the configuration file. A better implementation is we introduce a new `processor` which is designed to finish only this feature. What do you think? Could you spare time to finish this feature as a new `processor`?

llhhbc:
ok

PR #518, opened 2023-05-15
Fix null point crash

llhhbc:
## Description

```c
Thread 12 (Thread 0x7fc4fe7fc700 (LWP 19462)):
#0  0x00007fc5446e74b9 in waitpid () from /lib64/libc.so.6
#1  0x00007fc544664f62 in do_system () from /lib64/libc.so.6
#2  0x00007fc544665311 in system () from /lib64/libc.so.6
#3  0x00007fc545183edf in print_core (signum=11, info=0xc000f89bf0,
    secret=0xc000f89ac0, oldact=<optimized out>)
    at /kindling/probe/src/cgo/catch_sig.cpp:37
#4  <signal handler called>
#5  0x0000000000ab7e20 in github.com/Kindling-project/kindling/collector/pkg/model.(*AttributeMap).ToStringMap ()
#6  0x000000c005118eb8 in ?? ()
#7  0x000000c0038f52c8 in ?? ()
#8  0x0000000000874de5 in (*jsonEncoder).safeAddString ()
#9  0x000000c002fb2000 in ?? ()
#10 0x0000000001ae0d73 in type:* ()
#11 0x000000c0004d3d00 in ?? ()
#12 0x000000c0023f6e40 in ?? ()
#13 0x0000000000788a22 in encoding/json.(*Encoder).Encode ()
#14 0x000000c000d8e0a0 in ?? ()
#15 0x0000000000000006 in ?? ()
#16 0x000000000000015b in ?? ()
#17 0x0000000001a808a0 in type:* ()
#18 0x00007fc5140f3278 in ?? ()
#19 0x000000000040d12d in runtime.getitab ()
#20 0x0000000000000050 in ?? ()
#21 0x000000c000f80000 in ?? ()
#22 0x000000c003bcd4f0 in ?? ()
#23 0x000000c0038f5348 in ?? ()
#24 0x0000000000874897 in (*jsonEncoder).addKey ()
#25 0x000000c002fb2000 in ?? ()
#26 0x0000000001b4bec1 in go:string.* ()
#27 0x000000c0038f5310 in ?? ()
#28 0x0000000000ab7ff9 in github.com/Kindling-project/kindling/collector/pkg/model.(*AttributeMap).String ()
#29 0x0000000000000000 in ?? ()
```

### DIFF
--- a/collector/pkg/model/attribute_map.go
+++ b/collector/pkg/model/attribute_map.go
@@ -144,6 +144,9 @@ func (a *AttributeMap) ClearAttributes() {
 
 func (a *AttributeMap) ToStringMap() map[string]string {
 	stringMap := make(map[string]string)
+	if a == nil {
+		return stringMap
+	}
 	for k, v := range a.values {
 		stringMap[k] = v.ToString()
 	}


dxsup:
Although this check is fair to make, I don't think it is the root cause of the issue. A nil of `*AttributeMap` shouldn't call this function.
What is the context of this error? Did you modify the code ever? Could you attach the crash stack printed by Go? It contains more valuable information to figure out the root cause.

llhhbc:
After me add this check, this error has gone.

dxsup:
Yes, it will fix the current null pointer issue. But if we know which line of code calls the `ToStringMap` method and cause this issue, we can know why the `*AttributeMap` is nil and prevent the issue from the source.

llhhbc:
我不确定是哪里导致的，堆栈中输出的信息比较有限。我其它地方也改了些代码，但更多的只是加了些日志输出，应该不影响。

dxsup (approved):

PR #558, opened 2023-08-11
Open-source probe analyzer handles multi-threading

YDMsama:
## Description

We've optimized the networkanalyzer component of the open-source probe analyzer. By leveraging Go's goroutines, we've separated event reception from processing, allowing them to run concurrently. The ConsumeEvent function now queues events, while a new ProcessEvents function, running in its own goroutine, handles processing. This enhancement boosts efficiency, especially under high event loads.

## How Has This Been Tested?

The compiled image can run normally locally and receive data.


dxsup:
We should stop the channel and the goroutine when shutting down the analyzer.

collector/pkg/component/analyzer/network/config.go
```diff
@@ -28,6 +29,7 @@ type Config struct {
 
 func NewDefaultConfig() *Config {
 	return &Config{
+		EventChannelSize:      100000,
```
dxsup, line 32: Add this option to the configuration files and set the default value 10000.

dxsup (approved):

PR #566, opened 2023-08-28
Add configuration option to ignore DNS RCODE 3 as errors

YDMsama:
## Description

This pull request introduces a new configuration option ignore_dns_rcode3_error that allows users to specify whether DNS responses with RCODE 3 (Name Error) should be treated as errors or not.

## Motivation and Context
In Kubernetes clusters using the ClusterFirst DNS policy, KubeDNS may return RCODE 3 for public domains, which are usually treated as errors. This behavior may lead to unnecessary alerts and logging, especially when users are aware of this behavior and do not consider it an error.

## How Has This Been Tested?

The compiled image runs normally.
After setting the ignore_dns_rcode3_error configuration to true and running the dig nonexistentsubdomain.dns3test.com command, the absence of trace metrics in Kindling indicates that the configuration change is effective and DNS queries with an RCODE of 3 are no longer considered errors.


hocktea214:
Please add TestCase for TestDnsProtocol() in network_analyzer_test

hocktea214 (approved):

dxsup (approved):

PR #580, opened 2023-09-18
Feature/add custom k8s metadata collector

NeJan2020:
- Added an extra application for collecting k8s metadata named metadata-provider,with an API that list/watchk8s metadata 

## Description
Old:
```
- k8s-apiserver <--(watch)-- agent1,agent2,...,agentx
```

With new application
```
- k8s-apiserver  <-- (watch) -- metadata-provider 
- metadata-provider <--(watch)-- agent1,agent2,...,agentx
```

## Motivation and Context
In large-scale cluster, many agents start at the same time, which directly stresses the k8s-apiserver. The capabilities and data provided by metadata-providers are much smaller than that provided by k8s-apiserver , which reduce flow pressure and avoid impact on the k8s-apiserver.

## How Has This Been Tested?
TestCase will be added soon


camera-front/metadata-provider.dockerfile
dxsup, line 1: Did you put this file in the wrong place?

collector/pkg/metadata/kubernetes/hostport_map.go
dxsup, line 1: Any reason for changes in this file?

NeJan2020: 'Sync.Map' is not suitable for serialization, but this hostPort information is required for initialization of the agents' k8sMetadataCache.

It might be possible to recalculate HostPortMap on the agent side using GloabelPodInfo, which also passed during initialization.

dxsup (approved):

